PR #10, opened 2018-04-11
Adding imports and main to first README example

arashout:
Hey I'm a rust newbie and it took awhile to get the first example to work. So I thought I would the imports you need to get going


arashout:
Oh! That's pretty cool, I did not realize that. Will create another PR doing it the right way

PR #15, opened 2018-04-12
fix #12: add Locator::id

### LINKED ISSUE #12: Add `Locator::Id`
It looks like this can use the `XPath` under the hood:

https://stackoverflow.com/questions/6844585/xpath-to-get-another-attribute-by-id-attribute

Possible solution:`"//a[@id = 'unique-id']"`

vitiral:
this fixes #12

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -218,6 +218,9 @@ pub enum Locator<'a> {
     /// Find an element matching the given CSS selector.
     Css(&'a str),
 
+    /// Find an element using the given `id`
+    Id(&'a str),
+
     /// Find a link element with the given link text.
     ///
     /// The text matching is exact.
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -234,6 +237,10 @@ impl<'a> Into<webdriver::command::LocatorParameters> for Locator<'a> {
                 using: webdriver::common::LocatorStrategy::CSSSelector,
                 value: s.to_string(),
             },
+            Locator::Id(s) => webdriver::command::LocatorParameters {
+                using: webdriver::common::LocatorStrategy::XPath,
+                value: format!("//*[@id=\"{}\"]", s),
```
jonhoo: I wonder if we need to do some escaping here.. I think it's probably fine for now, but with more advanced identifiers that could totally be the case.

vitiral:
I dont think advanced id's is a strong use case.

PR #21, opened 2018-04-25
Implement commands `SetWindowRect` and `GetWindowRect`

### LINKED ISSUE #19: Implement window size/position manipulation
This is done via [SetWindowRect](https://docs.rs/webdriver/0.35.0/webdriver/command/enum.WebDriverCommand.html#variant.SetWindowRect).

The Selenium implementation provides [`set_window_size(width, height)`](https://github.com/SeleniumHQ/selenium/blob/38aa3268af0682973e46cd262aaa74d36c39415b/py/selenium/webdriver/remote/webdriver.py#L1054) and [`set_window_position(x, y)`](https://github.com/SeleniumHQ/selenium/blob/38aa3268af0682973e46cd262aaa74d36c39415b/py/selenium/webdriver/remote/webdriver.py#L1095) so we should probably provide those as well.

Phrohdoh:
If merged this will close #19.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -514,6 +514,8 @@ impl Client {
             WebDriverCommand::ElementSendKeys(ref we, _) => {
                 base.join(&format!("element/{}/value", we.id))
             }
+            WebDriverCommand::SetWindowRect(..) => base.join("window/rect"),
+            WebDriverCommand::GetWindowRect => base.join("window/rect"),
             _ => unimplemented!(),
         }
     }
@@ -571,6 +573,10 @@ impl Client {
             | WebDriverCommand::GoBack
             | WebDriverCommand::Refresh => {
                 body = Some("{}".to_string());
+                method = Method::Post;
+            }
+            WebDriverCommand::SetWindowRect(ref params) => {
+                body = Some(format!("{}", params.to_json()));
                 method = Method::Post;
             }
             _ => {}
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -777,6 +783,141 @@ impl Client {
...
     }
 
+    /// Sets the x, y, width, and height properties of the current window. All values must be `>= 0`.
+    pub fn set_window_rect(&self, x: i32, y: i32, width: i32, height: i32) -> impl Future<Item = Self, Error = error::CmdError> + 'static {
+        use webdriver::common::Nullable;
+
+        debug_assert!(x >= 0);
+        debug_assert!(y >= 0);
+        debug_assert!(width >= 0);
+        debug_assert!(height >= 0);
```
jonhoo: I think I'd prefer if these were errors rather than assertions.
Alternatively, make the arguments here `u32`s so that that can't happen, and the convert them to `i32` for the internal representation.

src/lib.rs
```diff
@@ -777,6 +783,141 @@ impl Client {
...
+    }
+
+    /// Gets the x, y, width, and height properties of the current window.
+    pub fn get_window_rect(&self) -> impl Future<Item = (i32, i32, i32, i32), Error = error::CmdError> + 'static {
+        self.dup()
+            .issue_wd_cmd(WebDriverCommand::GetWindowRect)
+            .and_then(|(_, v)| match v {
+                Json::Object(obj) => {
+                    let x = match obj.get("x") {
+                        Some(Json::U64(v)) => *v as i32,
```
jonhoo: Are these always `U64`? Is there not a generic "extract number" method on `Json` we can use?

src/lib.rs
```diff
@@ -777,6 +783,141 @@ impl Client {
...
+                _ => Err(error::CmdError::NotW3C(v))
+            })
+    }
+
+    /// Sets the width and height of the current window. All values must be `>= 0`.
+    pub fn set_window_size(&self, width: i32, height: i32) -> impl Future<Item = Self, Error = error::CmdError> + 'static {
+        use webdriver::common::Nullable;
+
+        debug_assert!(width >= 0);
+        debug_assert!(height >= 0);
```
jonhoo: Same here -- either take `u32` or return an error on negative values.

src/lib.rs
jonhoo: Arguably we should move to [`failure`](https://github.com/withoutboats/failure) which would make some of this nicer to deal with, but I think for now the thing to do would be to add a variant to `CmdError` to signify invalid parameters. Might come in handy elsewhere too!

src/lib.rs
Phrohdoh: No the `Json` enum has helper methods to determine which type a value is but there is no generic "number" that matches the JS spec.

In practice these can/should be `F64` but in reality I'm always seeing `U64` values at this point in the code.

jonhoo: I wonder if the other case should return a `NotW3C` error then?

Phrohdoh: I tried that but `json` is partially-borrowed here.

Perhaps I need to construct a new `Json::Object` here.

I'll investigate tomorrow.

jonhoo: If you use `obj.remove` instead of `obj.get`, it should work. You can also use [`Json::as_i64`](https://docs.rs/rustc-serialize/0.3.24/rustc_serialize/json/enum.Json.html#method.as_u64) for the casting.

Phrohdoh:
@jonhoo This is ready for review.

jonhoo:
I think the `rustfmt` run unfortunately made GitHub mark some of my recent comments outdated :(

src/lib.rs
```diff
@@ -777,6 +783,186 @@ impl Client {
...
     }
 
+    /// Sets the x, y, width, and height properties of the current window. All values must be `>= 0` or you will get a `CmdError::InvalidArgument`.
+    pub fn set_window_rect(&self, x: i32, y: i32, width: i32, height: i32) -> impl Future<Item = Self, Error = error::CmdError> + 'static {
+        use webdriver::common::Nullable;
+
+        if x < 0 {
+            return future::Either::A(future::err(error::CmdError::InvalidArgument(
+                stringify!(x).into(),
+                format!("Expected to be `>= 0` but was `{}`", x)
```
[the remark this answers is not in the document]
Phrohdoh: Alright that's what I had previously but I change it to what it is currently.

> Invalid argument `x`: Expected to be `>= 0` but was `<val>`

I can certainly change it back though.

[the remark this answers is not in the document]
jonhoo: Ohhhh, I missed that. No, this seems fine :)

src/lib.rs
```diff
@@ -777,6 +783,186 @@ impl Client {
...
+    /// Gets the x, y, width, and height properties of the current window.
+    pub fn get_window_rect(&self) -> impl Future<Item = (u64, u64, u64, u64), Error = error::CmdError> + 'static {
+        self.dup()
+            .issue_wd_cmd(WebDriverCommand::GetWindowRect)
+            .and_then(|(_, v)| match v {
+                Json::Object(mut obj) => {
+                    let x = match obj.remove("x").and_then(|x| x.as_u64()) {
+                        Some(x) => x,
+                        None => return Err(error::CmdError::NotW3C(Json::Object(obj))),
+                    };
```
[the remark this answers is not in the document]
Phrohdoh: Yes all 4 properties are returned https://w3c.github.io/webdriver/webdriver-spec.html#dfn-window-rect.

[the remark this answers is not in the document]
jonhoo: Ah, and [crucially](https://drafts.csswg.org/cssom-view/#dom-window-screenx):

> The screenX attribute must return the x-coordinate, relative to the origin of the Web-exposed screen area, of the left of the client window as number of CSS pixels, **or zero if there is no such thing**.

(and the same for the other attributes)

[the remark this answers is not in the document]
Phrohdoh: Would you like us to special-case `0` due to that or keep this as-is for consumers to handle?

jonhoo:
And I assume the tests pass if you run them unignored locally (given that they're currently ignored)

Phrohdoh:
Yes the tests pass if you run them one at a time (by unignoring only a single test).

jonhoo:
Huh... Why would that be?

Phrohdoh:
The driver I am using (`chromedriver`, and I assume all others) only accepts a single connection at a time and `cargo` tests are ran in parallel.

jonhoo:
:tada:

Issue #6, opened 2018-01-07
Request: Implement `select_by_value` for `select` elements


Phrohdoh:
I am attempting to do this locally but am tripping over Rc.

Current code is in ~[a gist](https://gist.github.com/Phrohdoh/34715986f26d93ce2bcbde94ef795071) (which is likely to become a 404 link once I fix this issue).~ in the following comment.

Any help / input much appreciated!

Phrohdoh:
Status update:

I have resolved the Rc issue (I do not know if my approach is correct though):

```rust
pub struct Select<'c> {
    c: &'c Client,
    e: Element,
}

impl<'c> Select<'c> {
    /// TODO: docs
    pub fn from(c: &'c Client, e: Element) -> Self {
        Self { c, e }
    }

    /// TODO: docs
    pub fn select_by_value(self, value: &'c str) -> impl Future<Item = Client, Error = error::CmdError> + 'c {
        let locator = format!("option[value=\"{}\"]", value); // TODO: Escape `value`
        let locator = webdriver::command::LocatorParameters {
            using: webdriver::common::LocatorStrategy::CSSSelector,
            value: locator,
        };

        // TODO: Use FindELementElements
        let cmd = WebDriverCommand::FindElementElement(self.e.e, locator);
        self.c.dup().issue_wd_cmd(cmd).and_then(move |(c, r)| {
            // NOTE: I am working on this part now. Need to convert `r` to an element, I believe.
            Ok(c)
        })
    }
}
```

jonhoo:
I'm not entirely sure why the `is_selected` branch is necessary? Couldn't you always just click the option? I also wonder if we should just fetch all the options under the `select` and do the checking for value in code. It'll save us the escaping of the value in the CSS selector, at the cost fo fetching more options (which probably doesn't matter all that much). Thoughts?

Phrohdoh:
If you always click an option you may be deselecting an already clicked option in a multi-select dropdown.

I do like the sound of fetching all options up front and doing comparisons in our code. Performance evaluations needed of course but the idea doesn't hurt.

jonhoo:
Yeah, I think I'd accept something like this. I'm a little tight for time in the coming weeks, but would be happy to take a look at and mentor a PR!

Phrohdoh:
Unfortunately I have run out of time for this and have been moved onto other projects that currently do not require this functionality meaning this issue is up-for-grabs for someone else to take a stab at.

[...]

PR #24, opened 2018-04-28
Implement dropdown option selection by value

Phrohdoh:
If merged this will close https://github.com/jonhoo/fantoccini/issues/6.

---

Usage of `Element::select_by_value` is illustrated below:

```rust
fn select_product_type(c: Client) -> impl Future<Item = Client, Error = CmdError> {
    c.find(Locator::Css("select[name='productType']"))
        .and_then(|e| e.select_by_value("Dairy"))
}
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1450,6 +1450,24 @@ impl Element {
             })
             .and_then(|(this, href)| this.goto(href.as_str()).map(|this| this))
     }
+
+    /// Find and click an `option` child element by its `value` attribute.
+    pub fn select_by_value(
+        self,
+        value: &str
+    ) -> impl Future<Item = Client, Error = error::CmdError> {
+        let locator = format!("option[value='{}']", value);
+        let locator = webdriver::command::LocatorParameters {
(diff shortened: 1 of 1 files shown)


Phrohdoh:
After writing the "reason" section in the OP here I cannot recall why I implemented this on a new type instead of on the existing `Client`.

It would, presumably, be just as simple to implement and use `Client::select_by_value(e, "foo")` which is what I had in mind when I wrote #6.

Phrohdoh:
I have pushed the above-pasted implementation.
It is much better to start simple and become complex only as necessary.

---

I will update the OP.

---

Usage is as follows:

```rust
fn select_product_type(c: Client) -> impl Future<Item = Client, Error = CmdError> {
    c.find(Locator::Css("select[name='productType']"))
        .and_then(|e| e.select_by_value("Dairy"))
}
```

jonhoo:
I think eventually we will want to have separate structs for specific types of elements (e.g., `Select`) that all `Deref` into `Element`, which is where this method would go. But for now, I think this is the right thing to do.

PR #32, opened 2018-06-18
Add Form::set

### LINKED ISSUE #31: Set value of input without name attribute
Please add a method for setting the value of an `input` element that doesn't have a `name` attribute.

I'm trying to set the value of the `input` element whose `id` is `autocomplete2` on https://www.reservecalifornia.com/CaliforniaWebHome/Facilities/AdvanceSearch.aspx.  The element doesn't have a `name` attribute, so `Form::set_by_name` doesn't seem to work.  `Client::find` can find the element, but `Element` doesn't appear to have a method for setting its value.

kraai:
Fixes #31

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1522,14 +1522,12 @@ impl rustc_serialize::json::ToJson for Element {
 }
 
 impl Form {
-    /// Set the `value` of the given `field` in this form.
-    pub fn set_by_name<'s>(
+    /// Set the `value` of the given `locator` in this form.
+    pub fn set<'s>(
         &self,
-        field: &str,
+        locator: Locator,
         value: &'s str,
     ) -> impl Future<Item = Self, Error = error::CmdError> + 's {
-        let locator = format!("input[name='{}']", field);
-        let locator = Locator::Css(&locator);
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -1522,14 +1522,12 @@ impl rustc_serialize::json::ToJson for Element {
 }
 
 impl Form {
-    /// Set the `value` of the given `field` in this form.
-    pub fn set_by_name<'s>(
+    /// Set the `value` of the given `locator` in this form.
```
jonhoo: We should make this a little more helpful. Perhaps:

> Find a field using the given `locator` and set its value to `value`.

src/lib.rs
```diff
@@ -1562,6 +1560,17 @@ impl Form {
             })
     }
 
+    /// Set the `value` of the given `field` in this form.
```
jonhoo: Now that we're touching this code anyway, I think I'd like this to say:

> Find a form input with the given `name` and set its value to `value`.

jonhoo:
Great! I left two minor comments inline which it'd be great if you could also fix. If you feel up to it, I'd also really appreciate it if you could add a test (like the existing `set_by_name` one).

PR #33, opened 2018-07-01
Support finding multiple elements

crabtw:


src/lib.rs
```diff
@@ -1245,6 +1251,14 @@ impl Client {
         self.by(search.into())
     }
 
+    /// Find elements on the page.
+    pub fn find_all(
+        &self,
+        search: Locator,
+    ) -> impl Future<Item = Vec<Element>, Error = error::CmdError> + 'static {
+        self.by_all(search.into())
```
jonhoo: Since nothing else needs `by_all`, I'd just place the contents of that method in here :)

jonhoo:
Awesome! Could you please add a test or two for this feature?

crabtw:
I have moved `by_all` into `find_all` and added one test.

src/lib.rs
```diff
@@ -1981,6 +2015,22 @@ mod tests {
             })
     }
 
+    fn finds_all_inner<'a>(c: &'a Client) -> impl Future<Item = (), Error = error::CmdError> + 'a {
+        // go to the Wikipedia frontpage this time
+        c.goto("https://en.wikipedia.org/")
+            .and_then(move |_| c.find_all(Locator::Css("#p-interaction li")))
+            .and_then(move |es| {
+                assert_ne!(es.len(), 0);
```
jonhoo: I think I'd prefer a stronger assertion than this. If we want to future-proof the test for changes to wikipedia, maybe check that a few known items are in the list?

crabtw:
It now checks the first 4 items.

jonhoo:
Perfect, thank you!

PR #35, opened 2018-09-26
Explicitly mention how to get variable values with execute()

bramhaag:
Explicitly mention that `return` has to be used in the execute method to retrieve the value of a variable.


jonhoo:
The current phrasing makes it a little unclear whether `return` needs to be used in the Rust code or the JS code. Could it say "you must use `return` in the JavaScript code"?

bramhaag:
I've updated the documentation @jonhoo.

jonhoo:
Thanks!

PR #38, opened 2018-10-12
Adds in method for sending keys

jimberlage:
I was looking for the [Element Send Keys method](https://www.w3.org/TR/webdriver/#element-send-keys) and couldn't find it, so I've added it in if you want it.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -179,7 +179,7 @@ extern crate webdriver;
 use http::HttpTryFrom;
 use serde_json::Value as Json;
 use tokio::prelude::*;
-use webdriver::command::WebDriverCommand;
+use webdriver::command::{SendKeysParameters, WebDriverCommand};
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -1017,6 +1017,22 @@ impl Element {
         })
     }
 
+    /// Simulate the user sending keys to an element.
+    ///
+    /// Note that since this *may* result in navigation, we give up the handle to the element.
```
jonhoo: I think this should probably just take `&mut self` instead, and return `impl Future<Item = ()`, since sending keys is unlikely to cause navigation directly.

jimberlage: Sure, we can do that

jonhoo:
Awesome, thanks!

jonhoo:
Published in 0.11.4

PR #39, opened 2018-10-28
Add default empty first-match capabilities.

cstorey:
When trying to run `fantoccini` against a local `selenium-server` instance with the main example in the documentation (albeit at http://localhost:4444/wd/hub/ ), I'd see an error of the form: `First match w3c capabilities is zero length`. 

The spec at https://www.w3.org/TR/webdriver1/#processing-capabilities mentions that `firstMatch` capabilities must be non-empty if defined, so this PR makes that happen.

### DIFF
--- a/src/session.rs
+++ b/src/session.rs
@@ -369,7 +369,7 @@ impl Session {
 
             let session_config = webdriver::capabilities::SpecNewSessionParameters {
                 alwaysMatch: cap.clone(),
-                firstMatch: vec![],
+                firstMatch: vec![webdriver::capabilities::Capabilities::new()],
             };
             let spec = webdriver::command::NewSessionParameters::Spec(session_config);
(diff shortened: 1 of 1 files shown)


jonhoo:
Hmm, this seems like a bug in Mozilla's `webdriver` crate then, in that they shouldn't include `firstMatch` if empty. But your fix seems reasonable anyway, so I'll merge it! Thanks :+1:

cstorey:
I guess you'd need an option in serde to not serialize the field at all if it's empty. Either way, thanks.

jonhoo:
Yeah, though `serde` has support for that kind of filtering with attributes anyway. I guess they just didn't include it.. I'm doing a `webdriver` bump anyway for another bugfix and then I'll issue a new release!

PR #47, opened 2019-06-29
Resolve warning about deprecated trait objects without explicit `dyn`

rye:
The compiler produces a
> warning: trait objects without an explicit `dyn` are deprecated

as a result of the `warn(bare_trait_objects)` compiler message being enabled by default.

This commit simply accepts suggested syntax, using `dyn` in these three locations where warnings are produced. I assume these suggestions are what was already being done internally by the compiler.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -32,7 +32,7 @@ impl Error for NewSessionError {
         }
     }
 
-    fn cause(&self) -> Option<&Error> {
+    fn cause(&self) -> Option<&dyn Error> {
         match *self {
             NewSessionError::BadWebdriverUrl(ref e) => Some(e),
             NewSessionError::Failed(ref e) => Some(e),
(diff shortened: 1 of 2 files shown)


jonhoo:
Ah, excellent, thank you! I'll merge once CI passes.

jonhoo:
I _think_ the failure is due to an outdated Ubuntu package. I'll re-run it a little bit later :)

jonhoo:
Apparently this is a known issue: https://travis-ci.community/t/how-to-setup-chromedriver-74-with-chrome-74-for-travis/2678/3

jonhoo:
https://github.com/jonhoo/fantoccini/commit/ff99c0dc41666cc9124771a553787514fa006021 should fix the CI issue. Can you rebase so we can check CI just for safety?

rye:
Absolutely! Just a moment. Thanks for hunting the issue down!

jonhoo:
Thanks for the fix!

PR #50, opened 2019-07-16
add method for element clear command

jh0l:
Implement the element clear command supported by webdriver
https://docs.rs/webdriver/0.39.0/src/webdriver/command.rs.html#270
https://w3c.github.io/webdriver/#element-clear
Basically renames the element click functionality

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1030,15 +1030,13 @@ impl Element {
         })
     }
 
-    /// Clear the contents of this element and then return that element
-    pub fn clear(self) -> impl Future<Item = Client, Error = error::CmdError> {
-        let e = self.e;
-        let mut c = self.c;
-        let cmd = WebDriverCommand::ElementClear(e);
-        c.issue(cmd).and_then(move |r| {
+    /// Clear the value prop of this element
+    pub fn clear(&mut self) -> impl Future<Item = (), Error = error::CmdError> {
+        let cmd = WebDriverCommand::ElementClear(self.e.clone());
+        self.c.issue(cmd).and_then(move |r| {
             if r.is_null() || r.as_object().map(|o| o.is_empty()).unwrap_or(false) {
                 // geckodriver returns {} :(
-                Ok(c)
+                Ok(())
             } else {
                 Err(error::CmdError::NotW3C(r))
             }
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -1030,6 +1030,19 @@ impl Element {
         })
     }
 
+    /// Clear the value prop of this element
+    pub fn clear(&mut self) -> impl Future<Item = (), Error = error::CmdError> {
+        let cmd = WebDriverCommand::ElementClear(self.e.clone());
+        self.c.issue(cmd).and_then(move |r| {
+            if r.is_null() || r.as_object().map(|o| o.is_empty()).unwrap_or(false) {
+                // geckodriver returns {} :(
```
jonhoo: This looks great! Could you just verify that this is indeed the case for `ElementClear` as well?

jh0l: yup so w3c web-driver api specifies for  the remote end of web-driver to 

> Return success with data null.

which is the same case for element click. The only thing I'm unsure of is how to handle errors for when the following happens
- If the current browsing context is no longer open, return error with error code no such window.
- Handle any user prompts and return its value if it is an error.
- Let element be the result of trying to get a known connected element with argument element id.
- If element is not editable, return an error with error code invalid element state.
- If element is not interactable, return error with error code element not interactable.

Right now it uses `error::CmdError::NotW3C`, will this work for all these types of errors?

jonhoo: I think ultimately we will want more semantic errors that capture more detailed error case like you lay out above, but I think `NotW3C` is fine for this PR. If the standard dictates the format of particular error messages returned by things like `ElementClear` then `NotW3C` is arguably somewhat misleading, but in general `issue` _should_ take care of any "standard" errors.

jonhoo: Also, what I actually originally was getting at here was whether the geckodriver workaround is needed for clearing elements too, or whether we can _just_ use the (standard) `is_null()`.

jh0l: Okay I'll check it out

jh0l:
Okay have cleaned up the unneeded code from Element::clear's response check, seems to work as intended.

jonhoo:
Excellent, thank you!

Issue #43, opened 2019-02-20
Add synchronous API wrapper


jonhoo:
As observed [here](https://www.reddit.com/r/rust/comments/aqgsxk/announcing_the_headless_chrome_crate_puppeteer/eghmzig/), there are some pretty good reasons to make the interface to interact with browsers synchronous. Or at least give the *option* of having the nicer synchronous interface when you don't need the async. We should be able to provide synchronous wrapper types for most operations pretty easily, so that would be a good addition!

lazypassion:
I'm new to async and not really that new to rust but still a beginner. How would I go about implementing this? Or what should I understand first?

jonhoo:
I think what we'd want here (if we even want this any more with async/await making async code so much easier to write) is a type that wraps an async runtime (tokio probably) + a fantoccini client, and then has synchronous versions of every method on the fantoccini client that just immediately await the future returned by the async version.

lazypassion:
Is it something like this:
```rust
use tokio::runtime::current_thread::Runtime
use fantoccini

struct Client {
    rt: Runtime
    client: fantoccini::Client
}

impl Client {
    pub fn new(webdriver: &str) -> Client {
        Client {
            rt: Runtime::new().unwrap(),
            client: fantoccini::Client::new("webdriver path"),
    }

    pub fn find(&mut self, search: Locator) -> Element {
        let task = async {
            client.find(search).await
        }
        self.rt.block_on(task)
    }
}
```

I didn't add any error handling but is the idea right?

jonhoo:
Yup, that seems basically right!

lazypassion:
Do you prefer it as one commit or broken up into multiple commits?

jonhoo:
Multiple commits is best, but it's fine for you to just push as you go -- we can always edit after. Multiple commits is just far easier to review, because I can easily review only what has changed since last time.

PR #51, opened 2019-07-18
Add support for std::future and async/await

### LINKED ISSUE #46: Futures 0.3
Since futures will be stable in 1.36, and the latest futures crate version has already ```Updated to stabilized futures_api``` as per https://github.com/rust-lang-nursery/futures-rs/releases  I think it would be a big ergonomic upgrade for the users if Fantoccini updated to futures 0.3. 

This would allow us to use the nightly async await feature without having to convert between 0.1 and 0.3 before every await call. 

This would alleviate issues such as https://github.com/jonhoo/fantoccini/issues/43 

Is this change something Fantoccini would be willing to embrace?

jonhoo:
This won't be merged quite yet, as it's nightly-only, but it's here for people to look at if they want!

It's quite neat:

```rust
#[tokio::main]
async fn main() -> Result<(), fantoccini::error::CmdError> {
    let mut c = Client::new("http://localhost:4444").await.expect("failed to connect to WebDriver");

    // first, go to the Wikipedia page for Foobar
    c.goto("https://en.wikipedia.org/wiki/Foobar").await?;
    let url = c.current_url().await?;
    assert_eq!(url.as_ref(), "https://en.wikipedia.org/wiki/Foobar");

    // click "Foo (disambiguation)"
    c.find(Locator::Css(".mw-disambig")).await?.click().await?;

    // click "Foo Lake"
    c.find(Locator::LinkText("Foo Lake")).await?.click().await?;

    // see that we ended up in the right place
    let url = c.current_url().await?;
    assert_eq!(url.as_ref(), "https://en.wikipedia.org/wiki/Foo_Lake");

    // close the browser
    c.close().await
}
```

Fixes #46.


caphrim007:
@jonhoo 

would this also be relevant to #43 ? looks pretty similar insofar as one gets away from the chained functions and replaces them with `await?` syntax.

Love your work and youtube videos btw

jonhoo:
@caphrim007 Hmm, yes, arguably, though a sync interface _would_ mean the caller doesn't need to worry about a runtime and such. I'm not entirely sure. Arguably the original argument [from Reddit](https://www.reddit.com/r/rust/comments/aqgsxk/announcing_the_headless_chrome_crate_puppeteer/eghmzig/) still holds with `await`, even though it's not quite as bad now.

tiberiusferreira:
Thank you so much for this. Really appreciate your work.

jonhoo:
Now that `hyper` and `hyper-tls` also have alphas with async/await support out, I've gone ahead and published an alpha of `fantoccini` 0.12! It still requires nightly, but at least it should now be easier to get, and you can see the docs on [docs.rs](https://docs.rs/fantoccini/0.12.0-alpha.1) :D

jonhoo:
Oh, if anyone wants to take a stab at writing up a changelog file, that'd be awesome! Something along the lines of https://github.com/jonhoo/inferno/blob/master/CHANGELOG.md.

tiberiusferreira:
Thanks, this is absolutely awesome!

Em qua, 4 de set de 2019 19:40, Jon Gjengset <notifications@github.com>
escreveu:

tiberiusferreira:
Hello, I'm hitting 
```
Compiling fantoccini v0.12.0-alpha.1
error[E0599]: no method named `try_concat` found for type `hyper::body::body::Body` in the current scope
 ```

Is it normal?

**Comment by tiberiusferreira** [comment]
I git cloned the repo, switched to the async branch and did ```cargo run```. Still the same error. I'm not sure how it passed CI.

Do you have any idea what might be happening?

**Comment by tiberiusferreira** [comment]
After quite a lot of digging I managed to build it enabling streams in hyper:
```
hyper = {version = "0.13.0-alpha.1", features = ["unstable-stream"]}
```

jonhoo:
Thanks! Should be fixed now in `alpha.2` :)

Issue #48, opened 2019-07-08
Full example


lazypassion:
Although there is some example code on how to use the library there isn't a full example that covers the setup. For example I wasn't really aware I would have to add the futures library as well as tokio in order to use this. So I think a full working example would be nice.

jonhoo:
That's a great idea! Do you think you might be able to write one up and submit it as a pull request?

lazypassion:
Yes that's my problem. I'm not able to get anything working. 

Here is my Cargo.toml:
```toml
[package]
name = "test"
version = "0.1.0"
authors = ["lazypassion <25536767+lazypassion@users.noreply.github.com>"]
edition = "2018"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[dependencies]
fantoccini = "0.11.7"
tokio = "0.1"
futures = "0.1.14"
```
and main.rs:
```rust
use fantoccini::{Client, Locator};
use futures::future::Future;
use tokio;

fn main() {
    let client = Client::new("http://localhost:4444");

    tokio::run(
        client
            .map_err(|error| {
                unimplemented!("failed to connect to WebDriver: {:?}", error)
            })
            .and_then(|client| {
                client.goto("https://www.wikipedia.org/")
            })
            .map_err(|error| {
                panic!("a WebDriver command failed: {:?}", error);
            })
    );
}
```

I'm getting the error:
```rust
error[E0271]: type mismatch resolving `<futures::future::map_err::MapErr<futures::future::and_then::AndThen<futures::future::map_err::MapErr<impl futures::future::Future, [closure@src/main.rs:10:22: 12:14]>, impl futures::future::Future, [closure@src/main.rs:13:23: 15:14]>, [closure@src/main.rs:16:22: 18:14]> as futures::future::Future>::Item == ()`
 --> src/main.rs:8:5
  |
8 |     tokio::run(
  |     ^^^^^^^^^^ expected struct `fantoccini::session::Client`, found ()
  |
  = note: expected type `fantoccini::session::Client`
             found type `()`
  = note: required by `tokio::runtime::threadpool::run`

error: aborting due to previous error

For more information about this error, try `rustc --explain E0271`.
```

jonhoo:
Oh, I see. The reason for that error is that `tokio::run` expects the `Future` you give it to have no "value". Whereas all the `fantoccini` futures return `fantoccini::Client`. Try adding a `.map(|_| ())` at the end?

[...]

PR #52, opened 2019-07-22
Full example using rust lang website

lazypassion:
#48

### DIFF
--- a/examples/basic_setup/README.MD
+++ b/examples/basic_setup/README.MD
@@ -2,6 +2,19 @@
 
 This example assumes you have geckodriver or chromedriver listening at port 4444.
 
+You can start the webdriver instance by: 
+
+### geckodriver
+`geckodriver --port 4444`
+
+### chromedriver
+`chromedriver --port=4444`
+
+The above commands assume either driver is on your path otherwise, provide the entire path
+to the binary: 
+
+`path/to/webdriver/geckodriver -p 4444`
+
 ## To Run
 
 Call `cargo run` from the root of the `basic_setup` directory.
--- a/examples/basic_setup/README.MD
+++ b/examples/basic_setup/README.MD
@@ -1,3 +1,7 @@
 ## Setup
 
 This example assumes you have geckodriver or chromedriver listening at port 4444.
+
+## To Run
+
+Call `cargo run` from the root of the `basic_setup` directory.
--- a/examples/basic_setup/README.MD
+++ b/examples/basic_setup/README.MD
@@ -1,0 +1,3 @@
+## Setup
+
+This example assumes you have geckodriver or chromedriver listening at port 4444.
--- a/examples/basic_setup/src/main.rs
+++ b/examples/basic_setup/src/main.rs
@@ -1,0 +1,45 @@
+use fantoccini::error::CmdError;
+use fantoccini::{Client, Locator};
+use futures::future::Future;
+use tokio;
+
+fn main() {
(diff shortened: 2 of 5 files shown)


examples/basic_setup/src/main.rs
```diff
@@ -0,0 +1,45 @@
...
+    let get_started_button =
+        r#"//a[@class="button button-download ph4 mt0 w-100" and @href="/learn/get-started"]"#;
+    let try_without_installing_button =
+        r#"//a[@class="button button-secondary" and @href="https://play.rust-lang.org/"]"#;
+    let play_rust_lang_run_button = r#"//div[@class="segmented-button"]/button[1]"#;
+
+    let rust_lang = client
+        .map_err(|error| unimplemented!("failed to connect to WebDriver: {:?}", error))
+        .and_then(|client| client.goto("https://www.rust-lang.org/"))
+        .and_then(|client| client_wait(client, 3000))
```
jonhoo: I don't think you should need all these waits :)

examples/basic_setup/README.MD
```diff
@@ -0,0 +1,7 @@
+## Setup
+
+This example assumes you have geckodriver or chromedriver listening at port 4444.
```
jonhoo: Could you include the command-lines for running those tools on that port here? Would probably be useful to someone wanting to try it themselves.

examples/basic_setup/src/main.rs
lazypassion: Oh yeah I know it isn't necessary, but if I don't put it there to the example flys by. You're talking about the `client_wait` right?

jonhoo: Oh, they're there to let the user observe the browser changing? Might be worth commenting that here so that people don't start using this pattern everywhere in their own code!

lazypassion: How would I go about doing that. Should I warn them not to use it or suggest to wait till a native solution comes along?

jonhoo: I would just put a comment in the first time you use it:
```rust
        .and_then(|client| {
            // we add an artificial delay here so that the example is easier to watch
            // you do not need these delays in your own code
            client_wait(client, 3000))
        })
```

jonhoo:
Perfect, thank you! I think now the only thing really missing is hinting at the existence of the example file in the top-level `README` and the documentation. It _should_ be sufficient to update the top-level comment in `src/lib.rs` and then run:
```console
$ cargo readme > README.md
```

lazypassion:
Sorry I'm new to this, which top level comment?

jonhoo:
Ah, sorry! So, fantoccini has built-in documentation that you can see at [docs.rs](https://docs.rs/fantoccini/0.11/fantoccini/). That documentation is generated from (among other places) the giant comment that starts [here](https://github.com/jonhoo/fantoccini/blob/master/src/lib.rs#L1). I'm thinking that you could extend that perhaps near the bottom:
https://github.com/jonhoo/fantoccini/blob/e3796cf4fe2f146bfe779b69bede280b84aa4a4e/src/lib.rs#L156

With something like
```rust
//! For more examples, take a look at the `examples/` directory.
```

The repository's README file is rendered from that same documentation using [`cargo-readme`](https://github.com/livioribeiro/cargo-readme/). So, you should only have to modify `src/lib.rs` as outlined above, and then run `cargo readme > README.md` to update the `README.md` file :)

jonhoo:
Looks good, thanks! :+1:

PR #53, opened 2019-08-20
Print unreachable errors received from WebDriver

evdokimovs:
This PR adds printing unreachable errors which we can receive from WebDriver.

Sometimes WebDriver send errors which fantoccini don't expect and silently panic with `unreachable!()`. This don't give information about why fantoccini panic and doesn't allow to deal with it.

Also I refactored code a bit. Fields with one letter name do not give understanding to those who see the code for the first time about what they store. Also specifying PATCH version in `Cargo.toml` don't make sense because `cargo` will ignore it and automatically bump PATCH version of those crates.

I added crate [serial_test_derive] for speedup your integration tests. This crate can run tests marked with `#[serial]` on one thread and other tests on multi thread. Since "chromedriver" can run several sessions in parallel, it is more reasonable to run tests for it in parallel, and "geckodriver" - in one thread.

[serial_test_derive]: https://crates.io/crates/serial_test_derive


jonhoo:
Thank you! I'll leave some inline notes :)

Cargo.toml
```diff
@@ -21,15 +21,19 @@ travis-ci = { repository = "jonhoo/fantoccini" }
...
 
 [dependencies]
-webdriver = "0.39.0"
-url = "1.5.0"
+webdriver = "0.39"
+url = "1.5"
 serde = "1.0"
 serde_json = "1.0"
-futures = "0.1.14"
+futures = "0.1"
```
jonhoo: It's not quite true that this is a "meaningless change". The patch version _is_ taken into considering by cargo, in that it will never pull a version _older_ than the given version. This may be important for dependencies that have max-version constraints, and can also be used to indicate required features from new patch versions so that cargo knows it has to fetch a version newer than what's in the user's `Cargo.lock`.

src/lib.rs
```diff
@@ -249,23 +252,23 @@ pub use session::Client;
...
+    form: webdriver::common::WebElement,
 }
 
 impl Client {
     /// Create a new `Client` associated with a new WebDriver session on the server at the given
     /// URL.
     ///
     /// Calls `with_capabilities` with an empty capabilities list.
-    #[cfg_attr(feature = "cargo-clippy", allow(new_ret_no_self))]
+    #[cfg_attr(feature = "cargo-clippy", allow(clippy::new_ret_no_self))]
```
jonhoo: I believe this should now just say:
```rust
#[allow(clippy::new_ret_no_self)]
```

src/session.rs
```diff
@@ -187,12 +188,12 @@ impl Ongoing {
...
     rx: futures::sync::mpsc::UnboundedReceiver<Task>,
-    c: hyper::Client<hyper_tls::HttpsConnector<hyper::client::HttpConnector>, hyper::Body>,
+    client: hyper::Client<hyper_tls::HttpsConnector<hyper::client::HttpConnector>, hyper::Body>,
     wdb: url::Url,
     session: Option<String>,
-    legacy: bool,
+    is_legacy: bool,
     ua: Option<String>,
-    persist: bool,
+    is_persist: bool,
```
jonhoo: This should still be called `persist`, not `is_persist`.

jonhoo:
Overall the change looks good subject to the changes above! :)
I'll also work on merging this into #51 when it lands.

evdokimovs:
@jonhoo ,

And one small question. Why do you put integration tests in source code? :) Maybe it would be better to put them in the `tests` folder?

evdokimovs:
@jonhoo ,

Fixed

jonhoo:
Oh, that's just an artifact of how the project started out as something relatively tiny and private. Feel free to separate them out in this PR too!

evdokimovs:
@jonhoo ,

> Feel free to separate them out in this PR too!

Done

PR #54, opened 2019-08-22
first implementation of sync fantoccini

arthmis:
Starting implementation of synchronous client and element


lazypassion:
It would be nice to have a sync branch to merge into instead of the await branch

jonhoo:
I'm going to go ahead and close this. I don't think it's worth the effort to maintain a complete sync wrapper around all of fantoccini, and in practice doing things with async isn't anywhere near as hard today as it was in 2019 :)

Issue #55, opened 2019-08-31
Expose "Find Element From Element" WebDriver API call


Follpvosten:
I'm currently trying out fantoccini, and this is one (most likely easily fixable) issue I've found: When working with HTML elements that don't have any clear identifier attached, it can be kind of hard to impossible to find their child elements while still keeping context.

Example:

```html
<div class="table">
  <div class="row">
    <div class="col name-col">Thing 1</div>
    <div class="col action-col"><button>Some action for Thing 1</button></div>
  </div>
  <div class="row">
    <div class="col name-col">Thing 2</div>
    <div class="col action-col"><button>Some action for Thing 2</button></div>
  </div>
</div>
```

Now what I want is:

1. Find each `div.row`. That is easily done with `find_all`.
2. Now, get each row's `div.name-col` text and `div.action-col > button` while keeping their relation and an `Element` on the button so I can invoke events on it.

At the moment, I can't see how I would perform the second step without falling back to invoking jQuery or something and parsing the result, or relying on indices (which is never a good idea).  
I think the right WebDriver API call would be [Find Element From Element](https://w3c.github.io/webdriver/#dfn-find-element-from-element); the `webdriver` crate has it as `WebDriverCommand::FindElementElement`.

API idea: Most WebDriver clients expose this call by having a "find element" method on their web element type; that would be `fantoccini::Element::find` here (and `fantoccini::Element::find_all` for "Find Elements From Element" to stay consistent with the `Client` API), which would internally invoke something like `fantoccini::Client::find_from`/`fantoccini::Client::find_all_from`. Those could also be pub in fantoccini, they work like find/find_all, but take a `fantoccini::Element` as the first argument.

It would work like this:

```rust
let rows = client.find_all(Locator::Css("div.table > div.row")).await?;
let mut col_groups = Vec::new();
for row in rows.iter() {
    let name = row.find(Css("div.name-col")).await?.text().await?;
    let button = row.find(Css("div.action-col > button")).await?;
    col_groups.push((name, button));
}
// Now I can present the user with the options and they can choose what to click on
```

I would really like to contribute and implement this myself as well, but as I only want to work with the `.await` API, I'd prefer to start working on it when the await PR is merged :)

jonhoo:
Hi! Ah, yes, this seems like a _great_ feature to add, and I agree that it's also something that's basically necessary for more advanced use-cases. You can start working from the `await` branch if you want (building on top of #51 that is) :)

[...]

PR #56, opened 2019-09-03
Add initial support for FindElement(s)FromElement

Follpvosten:
For #55 - really simple first draft by now, as most of the actual functionality was already implemented, I just needed to add an interface for it. I've updated src/session.rs in the process to support the latest hyper master (they seem to have removed that parameter).

At first, I thought about passing an `&Element` to the `find_from`/`find_all_from` methods, but that would be difficult as the `Element`'s `find`/`find_from` methods must take `&mut self` to be able to call the inner `Client`, but by passing on `&self` into the `find_from` Future, they would also return an immutable borrow of self, so I went to just cloning the `Element`.

Two questions:
1. Should I remove the `#![feature(async_await)]` from lib.rs and tests/common.rs? As it's stabilized on nightly now, these raise a warning
2. Any input on how I should test this? I've tried it out locally in my own project, and it worked just as expected, now I'm thinking about maybe adding a simple pre-baked HTML file to the tests and opening that in the browsers, as I can't really rely on Wikipedia's HTML staying the same forever. Some simple test ideas would be "find all links inside two spans/divs" or "find `ul` elements and list their `li`s".

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -124,7 +124,6 @@
 //! [WebDriver compatible]: https://github.com/Fyrd/caniuse/issues/2757#issuecomment-304529217
 //! [`geckodriver`]: https://github.com/mozilla/geckodriver
 #![deny(missing_docs)]
-#![feature(async_await)]
 
 use http::HttpTryFrom;
 use serde_json::Value as Json;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -897,7 +897,7 @@ impl Element {
         Ok(self.prop(prop).await?.unwrap())
     }
 
-    /// Find the first matching child element.
+    /// Find the first matching descendant element.
     pub async fn find(&mut self, search: Locator<'_>) -> Result<Element, error::CmdError> {
         let res = self
             .client
@@ -912,7 +912,7 @@ impl Element {
             element: e,
         })
     }
-    /// Find all matching child elements.
+    /// Find all matching descendant elements.
     pub async fn find_all(&mut self, search: Locator<'_>) -> Result<Vec<Element>, error::CmdError> {
         let res = self
             .client
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -673,41 +673,10 @@ impl Client {
         self.by(search.into()).await
     }
 
-    /// Find an element on the page, starting from the given element.
-    pub async fn find_from(
-        &mut self,
-        element: Element,
-        search: Locator<'_>,
-    ) -> Result<Element, error::CmdError> {
-        self.by_from(element, search.into()).await
-    }
-
     /// Find elements on the page.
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -778,6 +809,21 @@ impl Client {
         })
     }
 
+    async fn by_from(
```
jonhoo: I'm not sure I follow this name?

src/lib.rs
```diff
@@ -674,6 +674,15 @@ impl Client {
         self.by(search.into()).await
     }
 
+    /// Find an element on the page, starting from the given element.
+    pub async fn find_from(
```
jonhoo: I wonder if we should _just_ have the methods on `Element`, and not these "nested find" methods on `Client`. That would also remove the need for the `clone` as you could operate entirely on `&mut Element`.

jonhoo:
1. Yes please!
2. I've been using Wikipedia for the tests so far, and the relevant HTML there hasn't changed since I initially published this project :p That said, it would probably be better to just bundle some HTML files for all the tests. Maybe do that in a separate PR though? Ideally we'd transition all the tests to something that operated on local files.

src/lib.rs
Follpvosten: Ah, sorry, I basically just copied the whole code path for `find`, which uses `Client::by`. Seeing as `by_from` is not used anywhere else, I can just move the code into `find_from`.

src/lib.rs
Follpvosten: Sounds like it would make sense; I'll try to do it that way!

Follpvosten:
Yup, a separate PR makes sense. I'll add some Wikipedia tests to this one then.

src/lib.rs
Follpvosten: `Element::find` also makes more sense semantically, but I don't think we can really get rid of the clone, as a `WebDriverCommand::FindElementElement` requires an owned `WebElement`.

Follpvosten:
Does this look good?

src/lib.rs
```diff
@@ -898,6 +897,40 @@ impl Element {
         Ok(self.prop(prop).await?.unwrap())
     }
 
+    /// Find the first matching child element.
```
jonhoo: We probably shouldn't refer to "child elements", but rather something like "descendants", as `FindElementElement` is not restricted to immediate children.

jonhoo:
This looks good, thanks! Now just some tests and the one comment above about the docs :)

Follpvosten:
I'm not sure if the test is sufficient, I just wanted to push a first working draft before going to sleep :) I hope everything else is looking fine now!

jonhoo:
This looks great, thanks!

PR #63, opened 2019-11-04
Improvements to `wait_<…>` functions

dljsjr:
The `wait_for_find(…)` method on `Client` was taking ownership of the client instance (used `mut self` in the signature instead of a mutable reference to self).

This just changes it to a mutable reference so that calling wait_for_find doesn't move the client.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -692,13 +692,13 @@ impl Client {
     /// While this currently just spins and yields, it may be more efficient than this in the
     /// future. In particular, in time, it may only run `is_ready` again when an event occurs on
     /// the page.
-    pub async fn wait_for<F, FF>(&mut self, mut is_ready: F) -> Result<&mut Self, error::CmdError>
+    pub async fn wait_for<F, FF>(&mut self, mut is_ready: F) -> Result<(), error::CmdError>
     where
         F: FnMut(&mut Client) -> FF,
         FF: Future<Output = Result<bool, error::CmdError>>,
     {
         while !is_ready(self).await? {}
-        Ok(self)
+        Ok(())
     }
 
     /// Wait for the given element to be present on the page.
@@ -732,7 +732,7 @@ impl Client {
     pub async fn wait_for_navigation(
         &mut self,
         current: Option<url::Url>,
-    ) -> Result<&mut Self, error::CmdError> {
+    ) -> Result<(), error::CmdError> {
         let current = match current {
             Some(current) => current,
             None => self.current_url_().await?,
@@ -743,7 +743,9 @@ impl Client {
             let current = current.clone();
             // TODO: and this one too
             let mut c = c.clone();
-            async move { Ok(c.current_url().await? == current) }
+            async move {
+                Ok(c.current_url().await? != current)
+            }
         })
         .await
     }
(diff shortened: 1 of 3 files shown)


dljsjr:
There is something similar with `raw_client_for` and `with_raw_client_for` but I guess I wanted to see what happened with this PR first, I may not understand the design intent. But with those methods taking ownership of the Client it causes problems when trying to work with shared references like `Arc<Mutex<Client>>` in some situations like `FnMut` closures.

jonhoo:
Ah, so, my (perhaps erroneous) thinking here was that if you are waiting on something, you can't _also_ be doing other things. But `&mut self` conveys that same information just fine. Seems like a good change to me! Feel free to go ahead and change the other methods too. They should also no longer return `Self` then (luckily we're still in `alpha`!)

dljsjr:
It only really came up when I tried to move an `Arc<Mutex<…>>` clone in to an `FnMut` closure because I was working with Streams, at which point the move semantics caused the borrow checker to yell at me. I'll add those other changes and update the PR.

dljsjr:
PR wasn't updating so I switched base branches to trick it… 🤷‍♂

jonhoo:
I'm guessing you'll also have to update some tests to match this new change.

dljsjr:
The Travis job passed, are there any tests in particular I should look at?

jonhoo:
I guess we don't really have tests for these -- that's awkward :sweat_smile: 
No, I think you're fine. Though could you please change the other `wait_` methods to also take `&mut self`?

dljsjr:
Absolutely! I might even try to write a few tests...

jonhoo:
I think all the methods that take `&mut self` also no longer need to return `Self`.

dljsjr:
Question about `wait_for_navigation` as the intent from the code doesn't really match the doc comments.

The doc comments say that it waits for the client to navigate to a new URL, and it takes in the current URL as its argument. So my assumption is that the function should be monitoring the current URL and waiting for it to be dissimilar from the provided argument.

But the function body compares the client's current URL to the argument `==`.

There's clearly a bug here (I'm writing a test) but I'm wondering if the bug is in the function definition (should it be `!=`) or if the bug is in the documentation.

dljsjr:
RE: returning `&mut Self`, I thought it would be nice to still be able to use combinators on the wait functions.

jonhoo:
Ah, yes, that should be `!=`! Good catch.
The reason I'd like to move away from returning `Self` is that we don't do it elsewhere. I feel like we should either do it everywhere or nowhere..

dljsjr:
Gotcha, makes sense.

dljsjr:
Added some test cases! Wasn't quite sure how to make a good case for wait_for_navigation without a multithreaded executor being available and I know threading in a test environment is usually sketchy so I hacked it by targeting a known redirect with a `Client::goto()`. I would say that the test probably isn't super reliable in the sense that it'll definitely always pass but won't always actually "wait" should the redirect happen too quickly.

jonhoo:
I wonder if the trick here is going to be to just include an HTML file directly in the source tree that has a delayed redirect (using a `<meta>` tag probably). Then you could just point the test at that?

dljsjr:
Hmm, I hadn't thought about that. That could work.

dljsjr:
Just pushed a passing version of the test that uses an in-tree HTML file with a redirect attribute in the META tag.

jonhoo:
Excellent, thank you!

PR #67, opened 2020-01-06
Added example with async await

kubastick:


jonhoo:
Ah, that's a great idea, thanks!

Issue #70, opened 2020-01-30
Request: Add ability to switch frames


hgzimmerman:
I'm trying to test a webpage that makes use of an iframe using this library.
Unfortunately, in order to interact with elements inside of iframes, the Client needs to switch into the iframe.
There exists command enums in the webdriver crate for performing the necessary actions, so this should be possible:
* `WebDriverCommand::SwitchToFrame`
* `WebDriverCommand::SwitchToParentFrame`

I have tried to do this myself in my fork, but I could not get my changes to work, with errors I don't quite recall being returned. 
The PR with the relevant changes is available here: https://github.com/hgzimmerman/fantoccini/pull/1/files#diff-b4aea3e418ccdb71239b96952d9cddb6R765 (due to vendoring and rust version lock at work, I am basing this off of the v0.11.9 release, but the changes shouldn't be materially different against master).

This very well may be a bug in the webdriver crate, but since they lack a github page for the library, I'm filing the issue here with the hope that someone more familiar with that library may be able to assist.

firefox version: Mozilla Firefox 72.0.1
geckodriver version: geckodriver 0.26.0
OS version: Ubuntu 18.04.3 LTS
library version: v0.11.9 (fork)

EDIT:
This is the error following calling any of my `frame` functions:
```
WebDriverErr(Lost(Custom { kind: BrokenPipe, error: "WebDriver session was closed while waiting"}))
```

jonhoo:
Hmm, that looks an awful lot like the webdriver _server_ crashed. Could you try running `geckodriver` with debug output and see if it prints anything helpful?

hgzimmerman:
Given the following output, it doesn't look like it crashes:

```
...
1580416571084	Marionette	DEBUG	0 -> [0,10,"WebDriver:ElementClick",{"id":"02744e0f-9292-440f-a8cc-c3bdf9e9d6eb"}]
1580416571096	Marionette	TRACE	[15] Received DOM event beforeunload for http://localhost:9001/login
1580416571103	Marionette	TRACE	[15] Received DOM event pagehide for http://localhost:9001/login
1580416571104	Marionette	TRACE	[15] Received DOM event unload for http://localhost:9001/login
1580416571133	Marionette	TRACE	[15] Received DOM event DOMContentLoaded for http://localhost:9001/login_mfa
1580416571173	Marionette	TRACE	[15] Received DOM event pageshow for http://localhost:9001/login_mfa
[...]

PR #71, opened 2020-01-30
Add support for switching between frames

hgzimmerman:
closes https://github.com/jonhoo/fantoccini/issues/70

Adds support for switching into and out of iframes.

This currently is known to not work.
A minimal reproduction is forthcoming, but may take some time.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -674,7 +674,6 @@ impl Client {
         Ok(self)
     }
 
-
     /// Switches to the parent of the frame the client is currently contained within.
     pub async fn parent_frame(mut self) -> Result<Client, error::CmdError> {
         self.issue(WebDriverCommand::SwitchToParentFrame).await?;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -126,8 +126,8 @@ use serde_json::Value as Json;
 use std::convert::TryFrom;
 use std::future::Future;
 use tokio::sync::oneshot;
-use webdriver::command::{SendKeysParameters, WebDriverCommand};
-use webdriver::common::ELEMENT_KEY;
+use webdriver::command::{SendKeysParameters, WebDriverCommand, SwitchToFrameParameters};
+use webdriver::common::{ELEMENT_KEY, FrameId};
 use webdriver::error::WebDriverError;
 
 macro_rules! via_json {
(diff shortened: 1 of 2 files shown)


hgzimmerman:
Reproduction repo is here: https://github.com/hgzimmerman/fantoccini_iframe_example

jonhoo:
Hmm, the panic seems to point at
https://github.com/jonhoo/fantoccini/blob/9d4c77c723f3a1e8dda8039f0179f8254664c490/src/session.rs#L487
You'll have to add a match arm to that for `SwitchToParentFrame`.

hgzimmerman:
Thanks, I didn't look around a whole lot and figured that that detail was handled by the webdriver crate.

Some progress is made, with the client still crashing, but now with the following error:
```
thread 'tokio-runtime-worker' panicked at 'gai background task failed: JoinError::Cancelled', /home/hzimmerm/.cargo/registry/src/github.com-1ecc6299db9ec823/hyper-0.13.2/src/client/connect/dns.rs:144:30
```

and the last lines of geckodriver being:
```
1580493548082	Marionette	DEBUG	0 -> [0,4,"WebDriver:FindElement",{"using":"xpath","value":"//*[@id=\"inlineFrameExample\"]"}]
1580493548085	Marionette	DEBUG	0 <- [1,4,null,{"value":{"element-6066-11e4-a52e-4f735466cecf":"27e96fba-2d1a-4f7f-a91c-274e1e532cfe"}}]
1580493548086	webdriver::server	DEBUG	<- 200 OK {"value":{"element-6066-11e4-a52e-4f735466cecf":"27e96fba-2d1a-4f7f-a91c-274e1e532cfe"}}
1580493548086	webdriver::server	DEBUG	-> POST /session/cf7ae1ce-26fa-44ce-9381-07d5ccce5944/frame {"id":{"element-6066-11e4-a52e-4f735466cecf":"27e96fba-2d1a-4f7f-a91c-274e1e532cfe"}}
1580493548087	Marionette	DEBUG	0 -> [0,5,"WebDriver:SwitchToFrame",{"element":"27e96fba-2d1a-4f7f-a91c-274e1e532cfe"}]
1580493548088	Marionette	DEBUG	0 <- [1,5,null,{"value":null}]
1580493548089	webdriver::server	DEBUG	<- 200 OK {"value":null}
```

hgzimmerman:
Actually, upon re-running this, I can't reliably reproduce the above error, it only happens sometimes.

When that does happen, the geckodriver-attached firefox window doesn't close after the test completes, and the geckodriver instance needs to be killed and restarted. 

But, if I end my program with:
```rust
    c.close().await?;
    Ok(())
```
Then everything behaves as expected.

I think this is ok to merge, provided you are comfortable with that.

hgzimmerman:
Any update on this?

jonhoo:
Oh, sorry, I was under the impression that this was still broken somehow. But I suppose it's really just Geckodriver that doesn't like when a session goes away without being cleanly closed. Looks good to me!

jonhoo:
It _would_ be really good to have a test for this in the future though.

jonhoo:
Released on 0.12.1!

hgzimmerman:
Thanks! Sorry for the ambiguity about the correctness of the feature. I've been using it in my v0.11.9 fork for the last month and haven't seen any problems there. My demonstration app I made to test my changes also works reliably with the proper closing of the client connection and that your assessment of the client going away unexpectedly causing the crashes _appears_ to be correct.

I'd like to contribute a test for this, but I don't want to promise that I'll do it in a timely manner 😄. Finding a public site that is known to use iframes (and will continue to use iframes in the future) is probably the biggest impediment. 
It'd be a lot easier if you would be OK with me contributing a tiny webserver & static page that uses an iframe that points to some external page for the test. Like something along the lines of [my aforementioned example](https://github.com/hgzimmerman/fantoccini_iframe_example/blob/master/server/src/main.rs).

jonhoo:
I think including a small webserver for testing makes a lot of sense, and would be happy to accept something like that. It'd also make the other tests a lot less brittle!

PR #75, opened 2020-03-23
Tab and new window support

hgzimmerman:
Still a work in progress at the moment, but figured I should share.

I need the ability to swap tabs for some of the tests I'm writing, and so I've added support here to enable that.

I can attest that the added methods work based on my use of them in my work project, aside from `new_window`, which I have not had a need for yet.

I intend to contribute a local test suite as discussed earlier as part of this PR, but can't justify that effort at work, so I'll have to do it in my free time.


hgzimmerman:
Hey, I've started to work on a new way of running tests against HTML files served from a local server.

Quick question - I get `Session is already started` errors when running it as is after the first test in local runs.

The way I got this to work at work was to spawn a new geckodriver instance for each thread instead of having one running in the background. Do you think that is acceptable here?

jonhoo:
Ah, yes, it's because for some reason geckodriver doesn't like you driving multiple Firefox instances through one geckodriver instance. I have no idea why. This is why the Firefox tests are all marked as `#[serial]`. You'll probably want to copy that.

hgzimmerman:
I've updated the geckodriver in the CI setup from v0.24 to v0.26 (which isn't necessarily needed), but apparently Firefox 66 is required for this endpoint to be supported and the CI currently uses 63. [1]

[1] https://github.com/mozilla/geckodriver/releases/tag/v0.24.0

jonhoo:
It would be good to mention the `0.26` geckodriver requirement in the relevant methods.

hgzimmerman:
As far as I'm aware, per the patch notes, only the `new_window` method requires > v0.24 because it is the only one that relies on the `.../window/new` endpoint. Initial testing on v0.22 on my laptop before upgrading to v0.26 indicated that only this method failed to work.

It is documented here currently: https://github.com/jonhoo/fantoccini/pull/75/commits/6de1ea22154803423457ee7f5da2ce5196c5d16b#diff-b4aea3e418ccdb71239b96952d9cddb6R828

src/lib.rs
```diff
@@ -776,6 +779,88 @@ impl Client {
         })
     }
 
+    /// Gets the current window handle.
+    pub async fn get_window_handle(
```
jonhoo: How about we just call this `window`?

src/lib.rs
```diff
@@ -776,6 +779,88 @@ impl Client {
...
+    ) -> Result<webdriver::common::WebWindow, error::CmdError> {
+        let res = self.issue(WebDriverCommand::GetWindowHandle).await?;
+        match res {
+            Json::String(x) => Ok(webdriver::common::WebWindow(x)),
+            v => Err(error::CmdError::NotW3C(v)),
+        }
+    }
+
+    /// Gets a list of all active windows (and tabs)
+    pub async fn get_window_handles(
```
jonhoo: And this `windows`?

src/lib.rs
```diff
@@ -776,6 +779,88 @@ impl Client {
...
+    pub async fn switch_to_window(
+        &mut self,
+        window: webdriver::common::WebWindow,
+    ) -> Result<(), error::CmdError> {
+        let params = SwitchToWindowParameters { handle: window.0 };
+        let _res = self.issue(WebDriverCommand::SwitchToWindow(params)).await?;
+        Ok(())
+    }
+
+    /// May close the session if no other windows exist.
```
jonhoo: This could use some more documentation. What happens if other windows _do_ exist? Is this usually used to close _another_ window, and if so, should we provide a method that takes a `WebWindow` and closes it?

src/lib.rs
```diff
@@ -776,6 +779,88 @@ impl Client {
...
+    /// Creates a new window. If `is_tab` is `true`, then a tab will be created instead.
+    ///
+    /// Requires geckodriver > 0.24 and firefox > 66
+    ///
+    /// Windows are treated the same as tabs by the webdriver protocol.
+    /// The functions `new_window`, `switch_to_window`, `close_window`, `get_window_handle` and `get_window_handles`
+    /// all operate on both tabs and windows.
+    pub async fn new_window(
+        &mut self,
+        is_tab: bool,
```
jonhoo: Would this make more sense as `as_tab`, since the window has not yet been created?

tests/local.rs
```diff
@@ -0,0 +1,264 @@
+#[macro_use]
+extern crate serial_test_derive;
+extern crate fantoccini;
+extern crate futures_util;
+
+use fantoccini::{error, Client};
+use warp::Filter;
+use std::path::PathBuf;
+
+macro_rules! tester {
```
jonhoo: It'd be good if we could share most of this code between the remote and local testing code. Maybe in `tests/common.rs`?

tests/local.rs
```diff
@@ -0,0 +1,264 @@
...
+}
+
+lazy_static::lazy_static! {
+    static ref PORT_COUNTER: std::sync::atomic::AtomicU16 = std::sync::atomic::AtomicU16::new(8000);
+}
+
+
+fn setup_server() -> u16 {
+    let port: u16;
+    loop {
```
jonhoo: Why do this rather than just bind a `TcpListener` without specifying a port, and then extracting the auto-selected port from `listener.local_addr()`?

tests/local.rs
```diff
@@ -0,0 +1,264 @@
...
+    });
+    std::thread::sleep(std::time::Duration::from_secs(1));
+    port
+}
+
+
+
+/// Starts the fileserver
+async fn start_server(port: u16) {
+    let localhost = [0, 0, 0, 0];
```
jonhoo: You probably want [this instead](https://doc.rust-lang.org/std/net/struct.Ipv4Addr.html#associatedconstant.LOCALHOST).

tests/local.rs
```diff
@@ -0,0 +1,264 @@
...
+    c.close().await
+}
+
+async fn close_window(mut c: Client, port: u16) -> Result<(), error::CmdError> {
+    let url = sample_page_url(port);
+    c.goto(&url).await?;
+    c.new_window(true).await?;
+    let handles = c.get_window_handles().await?;
+    assert_eq!(handles.len(), 2);
+    c.close_window().await?;
```
jonhoo: It'd be good to have an assert after this that asserts what the _current_ window now is.

hgzimmerman: Good catch here. After closing a window, it doesn't switch to a new window automatically. Calling `window()` again after closing the window will cause the webdriver to return a 404, because the window it currently is "switched to" no longer exists.

Do you think the `window()` command should return a `<Result<Option<Window>, _>` or just throw an error like it currently does?

hgzimmerman: Scratch that - I've opted to keep the standard `Result<Window, _>` and made a new error variant to handle `NoSuchWindow` errors.

tests/local.rs
hgzimmerman:
```
    let port = TcpListener::bind("127.0.0.1").unwrap().local_addr().unwrap().port();
```
I just tried the above, and couldn't get it to work.
I'm inclined to leave it as is.

jonhoo: I would strongly prefer that we don't do port probing. Letting binding take care of port allocation is the standard way to do this, and much less error-prone. The trick here is probably going to be using [`warp::Server::bind_ephemeral`](https://docs.rs/warp/0.2.2/warp/struct.Server.html#method.bind_ephemeral), pass it a `SocketAddr` with the port set to `0`, and then the `SocketAddr` you get back has the port to use set.

tests/local.rs
jonhoo: I'm not sure what should happen here actually. In some sense I'm inclined to have `close_window` automatically switch to a different window if one exists, but that might be confusing too. `NoSuchWindow` seems pretty reasonable.

hgzimmerman: There's decent documentation for that behavior if lib users are inclined to read it.
Its a niche set of commands that aren't used frequently, given how I'm the first to need them. I think slightly idiosyncratic behavior is acceptable here, given that the alternative would be for the close method to also make commands to get other windows and switch to one of them.

tests/local.rs
hgzimmerman: I'm _really_ struggling with this approach. 
For whatever reason, starting the server in this manner causes it to not actually come up.
Usually this results in the first test failing to change the url to the expected address, because the server isn't present.

I've tried spinning for ~10 seconds with a http client to see if the server comes up, but it doesn't seem to. (firefox tests will fail if they can't get a valid response, while chrome tests seem to be ok with that. )

This is far more error prone than the former method, at least for the use case of short-lived test servers.

jonhoo: That is because you are constructing a `TcpListener`, getting its address, and then passing the address to `warp`. Instead, _just_ use warp's `bind_ephemeral` and pass 0 as the port, and then just extract the port number from the return value of `bind_ephemeral` :) You should not need any retries.

tests/local.rs
```diff
@@ -0,0 +1,351 @@
...
+use fantoccini::{error, Client};
+use std::net::{Ipv4Addr, TcpListener, SocketAddr};
+use std::path::PathBuf;
+use warp::Filter;
+use futures_core::Future;
+use fantoccini::error::CmdError;
+use warp::test::request;
+
+
+async fn select_client_type(s: &str) -> Result<Client, error::NewSessionError> {
```
jonhoo: Ah, sorry, what I meant by factor this out was to _also_ share it with the same code in `tests/common.rs` here:
https://github.com/jonhoo/fantoccini/blob/28087fb4e3ef5457c2400194c157a2818a46d435/tests/common.rs#L13

It seems unfortunate for us to have to do this complex matching + construction  in the same way in two different files.

tests/local.rs
hgzimmerman: Thanks. The latest change greatly simplified the server setup, but as far as I'm aware, it still doesn't work. Hopefully its easier to spot an error now.

tests/local.rs
hgzimmerman: I've moved the remote testing functions to another file and now moved all setup fns and macros to common.

jonhoo:
I like this latest change a lot! Leaving some more inline notes.

tests/common.rs
```diff
@@ -1,482 +1,160 @@
...
-    c.set_window_size(200, 100).await?;
-    c.set_window_position(0, 0).await?;
-    c.set_window_position(1, 2).await?;
-    let (x, y) = c.get_window_position().await?;
-    assert_eq!(x, 1);
-    assert_eq!(y, 2);
-
-    c.close().await
+#[macro_export]
+macro_rules! local_tester {
```
jonhoo: This and `tester!` are almost identical. We can probably merge them in some reasonable way. But if you're not familiar with macros, it can be a bit of a pain, so I can do that separately later :)

tests/common.rs
```diff
@@ -1,482 +1,160 @@
...
+/// Sets up the server and returns the port it bound to.
+pub fn setup_server() -> u16 {
+    let (tx, rx) = std::sync::mpsc::channel();
+
+    std::thread::spawn(move || {
+        let mut rt = tokio::runtime::Builder::new().enable_all().basic_scheduler().build().unwrap();
+        let _ = rt.block_on(async {
+            let (socket_addr, server) = start_server();
+            tx.send(socket_addr.port()).expect("To be able to send port");
+            server
```
jonhoo: Shouldn't this be `server.await`?

tests/common.rs
```diff
@@ -1,482 +1,160 @@
...
+        });
+    });
+    std::thread::sleep(std::time::Duration::from_secs(5));
```
jonhoo: This sleep should not be necessary.

tests/remote.rs
```diff
@@ -0,0 +1,409 @@
+#[macro_use]
```
jonhoo: I assume this file is mostly just a copy of the tests that used to be in `tests/common.rs`. Any other changes?

tests/local.rs
```diff
@@ -0,0 +1,183 @@
...
+async fn new_window_switch(mut c: Client, _port: u16) -> Result<(), error::CmdError> {
+    let handle_1 = c.window().await?;
+    c.new_window(false).await?;
+    let handle_2 = c.window().await?;
+    assert_eq!(
+        handle_1, handle_2,
+        "After creating a new window, the session should not have switched to it"
+    );
+
+    let all_handles = c.windows().await?;
```
jonhoo: Should we `assert_eq!(all_handles.len(), 2)` here?

tests/local.rs
```diff
@@ -0,0 +1,183 @@
...
+    let handle_3 = c.window().await?;
+    assert_ne!(
+        handle_3, handle_2,
+        "After switching to a new window, the handle should differ now."
+    );
+
+    c.close().await
+}
+
+async fn new_tab(mut c: Client, port: u16) -> Result<(), error::CmdError> {
```
jonhoo: Should we make this `new_tab_switch`, and have it test all the same things that `new_window_switch` does?

tests/common.rs
hgzimmerman: well... uuuh... that would explain a lot...

tests/remote.rs
hgzimmerman: It was just a plain copy out of  `tests/common`.

tests/common.rs
hgzimmerman: I tried earlier, but ran into problems with moving the port into an "async" closure so the test functions could be started in the same way because async closures aren't really a thing yet. I'm sure it can be done, but I'll hold off for now.

tests/local.rs
```diff
@@ -121,13 +141,13 @@ mod firefox {
     #[test]
     #[serial]
     fn new_window_switch_test() {
-        local_tester!(new_window_switch, "firefox")
+        tester!(new_window_switch, "firefox")
```
jonhoo: Why did these change to be non-local?

jonhoo:
Looks like you're missing some imports in `tests/remote.rs`: https://travis-ci.com/github/jonhoo/fantoccini/jobs/309413329#L1232

tests/local.rs
hgzimmerman: They don't need to be - they aren't using the local server for anything.
Ideally, the local server is used to test the recent Iframe additions, and possibly migrate some remote tests to, but the windowing tests can be done mostly without needing to access a given page.

hgzimmerman:
Ok, I've added a couple of tests unrelated to the named purpose of this PR to justify the effort of getting that server set up.

I'm happy where this is at currently, let me know if there is anything else you want me to change.

tests/local.rs
jonhoo: I actually wonder if ultimately we want to move almost _all_ the tests to be local, so that they won't accidentally break just because some website changes, but that's a change we can make later with the groundwork you've now laid down :)

tests/local.rs
```diff
@@ -28,7 +28,33 @@ async fn find_and_click_link(mut c: Client, port: u16) -> Result<(), error::CmdE
...
+    c.goto(&url).await?;
+    // Go to the page that holds the iframe
+    c.find(Locator::Css("#iframe_page_id")).await?.click().await?;
+
+    c.find(Locator::Id("iframe_button")).await.expect_err("should not find the button in the iframe");
+    c.find(Locator::Id("root_button")).await?; // Can find the button in the root context though.
+
+    // find and switch into the iframe
+    let iframe_element = c.find(Locator::Id("iframe")).await?;
+    iframe_element.frame().await?;
```
jonhoo: Reading this really makes me think that `frame()` should be renamed to `enter_frame()`. `frame()` makes me think that it's going to return a _handle_ to the frame, not switch to it.

tests/local.rs
```diff
@@ -28,7 +28,33 @@ async fn find_and_click_link(mut c: Client, port: u16) -> Result<(), error::CmdE
...
+
+    // search for something in the iframe
+    let button_in_iframe= c.find(Locator::Id("iframe_button")).await?;
+    button_in_iframe.click().await?;
+    c.find(Locator::Id("root_button")).await.expect_err("Should not be able to access content in the root context");
+
+    // switch back to the root context and access content there.
+    let mut c = c.parent_frame().await?;
```
jonhoo: And same here, `parent_frame` should probably be called `leave_frame()`. Or maybe `enter_parent_frame` is better?

jonhoo:
Happy to help! And thanks for sticking with it through all my feedback :) I think the only thing left now is to merge with master and resolve any conflicts, and then we're ready to release it!

jonhoo:
Do you agree with the iframe navigation renames? One downside of making them is that the change will require another major version bump, but :man_shrugging:

jonhoo:
Let me know if you have any trouble merging master. I don't think there are any major changes there that should cause you headache.

hgzimmerman:
I think I'm already based off the latest master. You should just be able to merge.

I agree with the naming changes to iframe functions, the fact that new names cause a version bump is your decision to make.

jonhoo:
Ah, it looks like GitHub was just confused and thought I wanted to rebase, never mind!

jonhoo:
@hgzimmerman I merged the macros in https://github.com/jonhoo/fantoccini/commit/d99623451c6ca72d35fe921d3baf70d558daf6e4 btw :)

tests/local.rs
jonhoo: Filed as https://github.com/jonhoo/fantoccini/issues/76

PR #77, opened 2020-04-02
Fix stackoverflow when printing CmdError

roblabla:
Printing a CmdError currently causes an infinite recursion, calling `<CmdError as Display>::fmt` over and over again until the stack inevitably overflows. This bug was introduced by https://github.com/jonhoo/fantoccini/commit/f05bd488d5dd0391003f61097843365145884730 . We basically undo this commit and add `#[allow(deprecated)]` instead. Those functions aren't going anywhere anyways, we're defining them ourselves!

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -173,8 +173,9 @@ impl Error for CmdError {
 }
 
 impl fmt::Display for CmdError {
+    #[allow(deprecated)]
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(f, "{}: ", self)?;
+        write!(f, "{}: ", self.description())?;
         match *self {
             CmdError::Standard(ref e)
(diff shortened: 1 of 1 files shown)


jonhoo:
I'm confused — won't this also recurse indefinitely, since `write!("{}", self)` ends up calling `Display`, which is what we're implementing? Could you add a test case to make sure we catch this going forward?

roblabla:
Erm, this PR changes `write!("{}", self)` to `write!("{}", self.description())`. The description function doesn't use display, it just returns a static string?

I'll add a test case.

roblabla:
d4d538d will fail with a stackoverflow on its own, the later two commits fix it.

jonhoo:
Oh, weird, the GitHub UI was showing me the inverted diff. No idea why that happened. The fix makes sense to me now :sweat_smile:

PR #87, opened 2020-05-09
Support the "stale element reference" error from WebDriver

ipkiss42:
This can happen when a handle on an element is kept but the element is
not on the page anymore (e.g. because it was removed from the DOM or
because the browser navigated to another page).

https://www.w3.org/TR/webdriver/#dfn-stale-element-reference

Note that I reused the existing `NoSuchElement` error variant. Let me know if you would prefer a new `StaleElementReference` error variant instead.

### DIFF
--- a/tests/local.rs
+++ b/tests/local.rs
@@ -285,7 +285,6 @@ mod chrome {
     }
 
     #[test]
-    #[serial]
     fn stale_element_test() {
         local_tester!(stale_element, "chrome")
     }
--- a/tests/local.rs
+++ b/tests/local.rs
@@ -166,6 +166,21 @@ async fn close_window_twice_errors(mut c: Client) -> Result<(), error::CmdError>
     Ok(())
 }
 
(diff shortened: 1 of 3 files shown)


jonhoo:
Ah, this is great, thank you! Could you please also add a test for this? The easiest way is probably to make the test use `Client::execute` to run some JavaScript that removes the element it just found.

ipkiss42:
Tests added. My first ones in Rust :)

tests/local.rs
```diff
@@ -258,4 +279,10 @@ mod chrome {
     fn double_close_window_test() {
         tester!(close_window_twice_errors, "chrome")
     }
+
+    #[test]
+    #[serial]
```
jonhoo, line 288: Ah, only the Firefox tests need `#[serial]` due to a bug in the Firefox WebDriver implementation. You do not need it here :)

jonhoo:
Great, thank you!

PR #89, opened 2020-05-11
Add json serialization of elements

### LINKED ISSUE #84: Element does not implement ToJson as advertized
The `Element::execute()` doc states:

```
args is available to the script inside the arguments array. Since Element implements ToJson,
you can also provide serialized Elements as arguments, and they will correctly serialize to DOM
elements on the other side.
```

However, `Element` doesn't seem to implement `ToJson`. See [this](https://github.com/jonhoo/fantoccini/search?q=ToJson&unscoped_q=ToJson) for example.

ipkiss42:
Closes #84.

### DIFF
--- a/tests/local.rs
+++ b/tests/local.rs
@@ -47,6 +47,15 @@ async fn serialize_element(mut c: Client, port: u16) -> Result<(), error::CmdErr
     )
     .await?;
 
+    // Check that it fails with an invalid serialization (from a previous run of the test)
+    let json = r#"{"element-6066-11e4-a52e-4f735466cecf":"fbe5004d-ec8b-4c7b-ad08-642c55d84505"}"#;
+    c.execute(
+        "arguments[0].scrollIntoView(true);",
+        vec![serde_json::from_str(json)?],
+    )
+    .await
+    .expect_err("Failure expected with an invalid ID");
+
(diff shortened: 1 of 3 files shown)


tests/local.rs
```diff
@@ -35,6 +35,21 @@ async fn find_and_click_link(mut c: Client, port: u16) -> Result<(), error::CmdE
...
+    let url = sample_page_url(port);
+    c.goto(&url).await?;
+    let elem = c.find(Locator::Css("#other_page_id")).await?;
+
+    // Check that webdriver understands it
+    c.execute(
+        "arguments[0].scrollIntoView(true);",
+        vec![serde_json::to_value(elem)?],
+    )
+    .await?;
```
jonhoo, line 48: Does this actually test that the serialization works correctly? Just as a sanity check, could you check that passing in some other random string here and see that the test indeed fails? I'm not sure whether the exception in the JavaScript code actually bubbles up to the WebDriver.

ipkiss42: Yes, it really tests it. I manually provided an incorrect element ID instead and got an error.

I could have left it in the code though. I will do it shortly.

jonhoo:
Excellent, thank you!

PR #92, opened 2020-07-17
Extensions feature

whizsid:
Ability to install or uninstall extensions and control browser specific features.

### DIFF
--- a/src/session.rs
+++ b/src/session.rs
@@ -10,7 +10,7 @@ use std::pin::Pin;
 use std::task::Context;
 use std::task::Poll;
 use tokio::sync::{mpsc, oneshot};
-use webdriver::command::{WebDriverCommand, WebDriverExtensionCommand};
+use webdriver::command::{WebDriverCommand, WebDriverExtensionCommand, VoidWebDriverExtensionCommand};
 use webdriver::error::ErrorStatus;
 use webdriver::error::WebDriverError;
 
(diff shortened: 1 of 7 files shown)


whizsid:
Some test cases failed. I think the websites has changed.
```
left: `["Help", "Community portal", "Recent changes", "Upload file"]`,
 right: `["Help", "About Wikipedia", "Community portal", "Recent changes"]`
```
And exceeded the maximum time.

src/session.rs
```diff
@@ -10,24 +10,48 @@ use std::pin::Pin;
...
 use webdriver::error::ErrorStatus;
 use webdriver::error::WebDriverError;
 
 type Ack = oneshot::Sender<Result<Json, error::CmdError>>;
 
 /// A WebDriver client tied to a single browser session.
 #[derive(Clone, Debug)]
-pub struct Client {
-    tx: mpsc::UnboundedSender<Task>,
+pub struct Client<T: ExtensionCommand> {
```
jonhoo, line 21: Making the `Client` generic over this is a _huge_ and backwards-incompatible change that I'd prefer to avoid. I think it'd be better to instead use a trait object, and pass around `Box<dyn WebDriverExtensionCommand>`. That way you can get rid of all these `<T>` :)

jonhoo:
Sorry it took me so long to get back to you!

src/session.rs
whizsid: @jonhoo I am getting an error when using `Box<dyn WebDriverExtensionCommand>`.  Error:- `the trait `WebDriverExtensionCommand` cannot be made into an object` . Because `WebDriverExtensionCommand` is extend to `PartialEq`. Can you help me to resolve this:- https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=91be6634f09c530866951321286b8d2c

jonhoo: Oh, hmm, that's really awkward... That seems like a bug in the `WebDriverExtensionCommand` trait to me, which will unfortunately require reporting a bug to that crate to resolve. Without that, there isn't a way for us to add this functionality without significantly complicating the API for (seemingly) no reason.

jonhoo: @whizsid did you end up filing an issue with the upstream crate for this?

whizsid: @jonhoo Yes. I submitted an issue. They shipped it with the latest release. I will update my PR.

ChaseCares:
Hello, I am interested in this feature for a project I am currently working on. @whizsid Do you have any interest in continuing with it?  I'm also willing to rework the traits, rebase and resubmit. Thanks!

PR #94, opened 2020-07-22
Add select_by_index method.

zhiburt:
Used to get an N option in select element.

Inspired by Java method

```javalistbox.selectByIndex(0);```

Probably there is a better implementation just using css selector or xpath but I didn't manage to get it.

The PR can be closed if it's considered useless method :)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1125,16 +1125,9 @@ impl Element {
     ///
     /// Index starts from 0.
     pub async fn select_by_index(mut self, index: usize) -> Result<Client, error::CmdError> {
-        let options = self.find_all(Locator::Css("option")).await?;
+        let locator = format!("option:nth-of-type({})", index + 1);
 
-        if options.len() > index {
-            options[index].clone().click().await
-        } else {
-            Err(error::CmdError::NoSuchElement(WebDriverError::new(
-                webdriver::error::ErrorStatus::NoSuchElement,
-                format!("there is no element with by {} index", index),
-            )))
-        }
+        self.find(Locator::Css(&locator)).await?.click().await
     }
 
     /// Switches to the frame contained within the element.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1123,7 +1123,11 @@ impl Element {
 
     /// Find and click an `option` child element by its index.
     ///
+    /// It expects there only to be one set of `option`s. 
     /// Index starts from 0.
+    ///
+    /// It uses a CSS selector `nth-of-type` therefore it ignores other's
+    /// type of elements except of an `option`.
     pub async fn select_by_index(mut self, index: usize) -> Result<Client, error::CmdError> {
         let locator = format!("option:nth-of-type({})", index + 1);
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1131,17 +1131,6 @@ impl Element {
     /// in the form, or if it there are stray `<option>` in the form.
     ///
     /// The indexing in this method is 0-based.
-    ///
-    /// It uses a CSS selector [nth-of-type] therefore it ignores other's
-    /// type of elements except of an `<option>`.
-    ///
-    /// To get a text value of a `<select>` element you can use [prop]
-    /// ```no_run
-    /// select_element.prop("value").await?
-    /// ```
-    ///
-    /// [prop]: Element::prop
-    /// [nth-of-type]: https://developer.mozilla.org/en-US/docs/Web/CSS/:nth-child
     pub async fn select_by_index(mut self, index: usize) -> Result<Client, error::CmdError> {
         let locator = format!("option:nth-of-type({})", index + 1);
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by jonhoo** [comment]
I think what you want here is probably [`:nth-of-type`](https://developer.mozilla.org/en-US/docs/Web/CSS/:nth-of-type). It's not _perfect_, because it only counts the index _among the element's siblings_, but for `option` it should be okay because (I believe) a `select` generally only contains `option`s. It would mean that you get weird results if you, say, call `select_by_index` on a `form` that has multiple `select`s (and thus multiple `option`s with a given index).

**Comment by zhiburt** [comment]
I'll follow up your advice.

I've tried once [`:nth-child`](https://developer.mozilla.org/en-US/docs/Web/CSS/:nth-child) way and I've also thought that it's possible by xpath `option[index]` but I didn't manage neither of them :(

**Comment by jonhoo** [comment]
It would also be good to add a test case for this! :)

**Comment by zhiburt** [comment]
I've tested only firefox version

```
[zhiburt@zhiburt fantoccini]$ cargo test --test local -- firefox::select_by_index_test --exact
   Compiling fantoccini v0.13.1 (/home/zhiburt/proj/fantoccini)
    Finished test [unoptimized + debuginfo] target(s) in 29.68s
     Running target/debug/deps/local-0c98d2dd68f4ea31

running 1 test
test firefox::select_by_index_test ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 21 filtered out
```

File: tests/local.rs [hunk 1]
```diff
@@ -209,6 +209,25 @@ async fn stale_element(mut c: Client, port: u16) -> Result<(), error::CmdError>
...
+    let initial_text = select_element.html(false).await?;
+
+    // Select second option
+    select_element.clone().select_by_index(1).await?;
+
+    // Get display text after selection
+    let text_after_selecting = select_element.html(false).await?;
+
+    // Make sure that it's changed
+    assert_ne!(initial_text, text_after_selecting);
```
jonhoo, line 227: It'd probably be good to assert that the text here is in fact the correct text. Maybe use `value` on the `select` itself to check that it indeed has the right option selected. Might also be a good idea to check that the other select did _not_ change.

src/lib.rs
```diff
@@ -1121,6 +1121,15 @@ impl Element {
         .await
     }
 
+    /// Find and click an `option` child element by its index.
+    ///
+    /// Index starts from 0.
```
jonhoo, line 1126: This is looking great now! I think the one thing I'd add is some more extensive documentation of this method. In particular, I think it should mention that it expects there to only be one set of `option`s, and that it uses the CSS selector `nth-of-type`.

zhiburt: > I think it should mention that it expects there to only be one set of options

Didn't clearly understand the idea so I basically copied that statement :)

jonhoo: Try this:
```rust
/// Find and click an `<option>` child element by its index.
///
/// This method clicks the first `<option>` element that is an `index`th child
/// (`option:nth-of-type(index+1)`). This will be the `index`th `<option>`
/// element if the current element is a `<select>`. If you use this method on
/// an `Element` that is _not_ a `<select>` (such as on a full `<form>`), it
/// may not do what you expect if there are multiple `<select>` elements
/// in the form, or if it there are stray `<option>` in the form.
///
/// The indexing in this method is 0-based.
```

zhiburt: I've decided to place there a straightforward example how to obtain text from `<select>` when it will be changed (in a way you've pointed me out yesterday) by `prop` since nor `html(true)` nor `text()` is suitable for this as I understand.

src/lib.rs
```diff
@@ -1121,6 +1121,33 @@ impl Element {
...
+    /// This method clicks the first `<option>` element that is an `index`th child
+    /// (`option:nth-of-type(index+1)`). This will be the `index`th `<option>`
+    /// element if the current element is a `<select>`. If you use this method on
+    /// an `Element` that is _not_ a `<select>` (such as on a full `<form>`), it
+    /// may not do what you expect if there are multiple `<select>` elements
+    /// in the form, or if it there are stray `<option>` in the form.
+    ///
+    /// The indexing in this method is 0-based.
+    ///
+    /// It uses a CSS selector [nth-of-type] therefore it ignores other's
```
jonhoo, line 1135: I don't think you need this particular paragraph given the text that's now over :)

src/lib.rs
```diff
@@ -1121,6 +1121,33 @@ impl Element {
...
+    /// type of elements except of an `<option>`.
+    ///
+    /// To get a text value of a `<select>` element you can use [prop]
```
jonhoo, line 1138: This isn't quite accurate — `prop("value")` gives you the value of the currently selected option element of the select. By default, that's the text contents of the `<option>`, but frequently it's not. For example:
```html
<option value="1">number one</option>
```

zhiburt: Now I see, it's worth while just to switch to your version completely :)

jonhoo:
This looks great, thank you!

jonhoo:
Published in 0.14 :tada:

Issue #27, opened 2018-05-29
Improve wait_* style methods


jonhoo:
Currently, `wait_for_find` [just spins](https://github.com/jonhoo/fantoccini/blob/ee0c585f6fb97c1718bd684055172a3ae067d02b/src/lib.rs#L1260) trying to look for the element in question. A much better strategy would be to use [`MutationObserver`](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) (specifically [`observe`](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver/observe) on the document root) to wait for new elements to be added to the page, and only re-execute the webdriver `find` when something new has appeared.

zhiburt:
Hey @jonhoo, what do you think about adding wrappers for `wait_*` functions with timeout parameter?
They possible can be done like this

```rust
    pub async fn wait_for_find_with_timeout(&mut self, search: Locator<'_>, timeout: Duration) -> Result<Element, error::CmdError> {
        tokio::time::timeout(timeout, self.wait_for_find(search))
    }
```

 As you've pointed out users can do it on there own but it might be helpful to have out of the box.

PS: Thanks for such a good crate

jonhoo:
So, I think wrappers like that actually add relatively little sadly. For that, the user can already just do the `tokio::time::timeout` call themselves. What I think I'd like to see is something a little more nuanced where, if the request times out, the WebDriver requests is cancelled. I don't know how feasible that is though? My concern is that the user issues a request with a timeout, it times out, then they issue another request, and that request fails or hangs for a long time because the browser already has a pending request...

zhiburt:
Could I be enlighten here since I am not proficient in webdriver architecture?:)

Do you tend to say that canceling requests by `tokio::time::timeout` might cause issues in the browser because it handles requests one by one or as if the browser handles it in a way that each request is handled in some `thread/goorutine/ etc.` the fast canceling of requests may cause slowing browser hugely since it will handle a ton of dead `thread/goorutine/ etc.`. That's your concern about such a way of timeout handling right?

jonhoo:
Yes, essentially. Although not necessarily that there are lots of dead requests, but rather that we only run one WebDriver request at the time currently. Imagine that you run a request that invokes an infinite loop in JavaScript. You time out the future that waits for that request, but the connection to the WebDriver instance is still held up by that (infinite) request. If you issue another fantoccini request, that request will forever be queued behind the still ongoing one.

PR #95, opened 2020-07-29
Create wait structure

zhiburt:
### Overview

Hello @jonhoo, I've tried to improve `*_wait` methods by creating a `Wait` structure which will allow to create it's own `rules`.

The design is primarily inspired by C# driver.
Here's an example how it's done in C#.

```c#
var wait = new WebDriverWait(driver, 100);
wait.Until(d => d.FindElements(By.Id("processing")).Count == 0);
```  

And here this example with new `Wait` structure. 

```rust
let mut wait = new Wait::new(client, timeout);
wait.until(|d| async { d.FindElement(Locator::Id("processing")).map_or(None, |_| Some(()) });
```

We also could create a set of predefined methods `wait_for_visible`, `wait_for_not_present` etc...

link #27 
link  #85 

### Implementation

Sadly I didn't figure out a way to support closures properly. There's a huge issue with lifetimes therefore in changed [`wait_for_find`](https://github.com/zhiburt/fantoccini/blob/a69c3370b53df95c3235828c30eb3e60e8527098/src/lib.rs#L732-L758) method I created a nested function which takes a cloned Client. That's only because of my incompetence here otherwise it had to be as simple as in example. I've not resolved the issue yet and decided to send the draft with believe that you may know some of pathways  to resolve this an issue :smile: .

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -146,6 +146,9 @@ pub use hyper::Method;
 /// Error types.
 pub mod error;
 
+/// Wait logic
+pub mod wait;
+
 /// The long-running session future we spawn for multiplexing onto a running WebDriver instance.
 mod session;
 use crate::session::{Cmd, Session};
@@ -711,7 +714,7 @@ impl Client {
     /// While this currently just spins and yields, it may be more efficient than this in the
     /// future. In particular, in time, it may only run `is_ready` again when an event occurs on
     /// the page.
-    pub async fn wait_for<F, FF>(&mut self, mut is_ready: F) -> Result<(), error::CmdError>
+    pub async fn wait_for<'a, F, FF>(&mut self, mut is_ready: F) -> Result<(), error::CmdError>
     where
         F: FnMut(&mut Client) -> FF,
         FF: Future<Output = Result<bool, error::CmdError>>,
@@ -727,20 +730,31 @@ impl Client {
     /// future. In particular, in time, it may only run `is_ready` again when an event occurs on
     /// the page.
     pub async fn wait_for_find(&mut self, search: Locator<'_>) -> Result<Element, error::CmdError> {
-        let s: webdriver::command::LocatorParameters = search.into();
-        loop {
-            match self
-                .by(webdriver::command::LocatorParameters {
-                    using: s.using,
-                    value: s.value.clone(),
-                })
-                .await
-            {
-                Ok(v) => break Ok(v),
-                Err(error::CmdError::NoSuchElement(_)) => {}
-                Err(e) => break Err(e),
+        fn closure(
+            s: Locator<'_>,
+            mut client: Client, // TODO: handle lifetime issues with &mut Client type
+        ) -> impl Future<Output = Option<Result<Element, error::CmdError>>> {
+            let s: webdriver::command::LocatorParameters = s.into();
+            async move {
+                match client
+                    .by(webdriver::command::LocatorParameters {
+                        using: s.using,
+                        value: s.value.clone(),
+                    })
+                    .await
+                {
+                    Ok(v) => Some(Ok(v)),
+                    Err(error::CmdError::NoSuchElement(_)) => None,
+                    Err(e) => Some(Err(e)),
+                }
             }
         }
+
+        const DEFAULT_TIMEOUT: std::time::Duration = std::time::Duration::from_millis(500);
+        let mut wait = wait::Wait::new(self, DEFAULT_TIMEOUT);
+        wait.until(move |client| closure(search, client.clone()))
+            .await
+            .map_err(|timeout| error::CmdError::Timeout(timeout))?
     }
 
     /// Wait for the page to navigate to a new URL before proceeding.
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -727,20 +730,31 @@ impl Client {
...
-                .by(webdriver::command::LocatorParameters {
-                    using: s.using,
-                    value: s.value.clone(),
-                })
-                .await
-            {
-                Ok(v) => break Ok(v),
-                Err(error::CmdError::NoSuchElement(_)) => {}
-                Err(e) => break Err(e),
+        fn closure(
```
jonhoo, line 733: I think your problem likely stems from trying to define a `fn closure` here. If you just put this code directly into the `move |client| {}` below, I think you should be fine, and not need the clone?

src/lib.rs
```diff
@@ -727,20 +730,31 @@ impl Client {
...
+                    .await
+                {
+                    Ok(v) => Some(Ok(v)),
+                    Err(error::CmdError::NoSuchElement(_)) => None,
+                    Err(e) => Some(Err(e)),
+                }
             }
         }
+
+        const DEFAULT_TIMEOUT: std::time::Duration = std::time::Duration::from_millis(500);
```
jonhoo, line 753: I think, by default, this method should never time out. The user asked us to wait indefinitely. If the user wants a timeout, they can do so by wrapping the returned future in a `tokio::time::timeout`.

src/wait.rs
```diff
@@ -0,0 +1,70 @@
...
+
+impl<'a> Wait<'a> {
+    ///
+    pub fn new(client: &'a mut Client, timeout: Duration) -> Self {
+        Self {
+            waiter: DefaultWait::new(client, timeout, DEFAULT_POOLING_INTERVAL),
+        }
+    }
+
+    pub(crate) async fn until<F, FF, R>(&mut self, condition: F) -> Result<R, Duration>
```
jonhoo, line 22: I think the whole `Wait` type only really makes sense if users can define their own waiters, which would mean making this method `pub`, no?

jonhoo:
Hopefully the comments above help a little! I'm still not quite sure what adding this kind of `wait` module actually helps with? As in, why is it better than the user just writing this loop themselves? The only reason I explicitly implemented `wait_for_find` and `wait_for_navigation` is that we can in theory be smarter about the loop in those two specific cases. For `wait_for_find`, we might be able to use a `MutationObserver` to only retry the find when the page changes, rather than on a timer. And similarly, for `wait_for_navigation`, we may be able to hook into a document onload event somehow. I don't think that's true for a "general" wait as implemented here, at which point it feels like the user can just add their own wait loop?

src/lib.rs
zhiburt: I hope that you're speaking about changes in `wait_for` here's you are right.
It's just a method which I chose to demonstrate the `Wait` :smile: and didn't noticed that I break it's logic :smile: 

Probably there could be introduced a new type WaitForever which would do just `loop`. And after we could amalgamate these by a trait `Waiter` which has a method `until`.

src/wait.rs
zhiburt: It was meant to be a `base` implementation on which could be piled different peaces. But I didn't decide at the time it created for users or for lib itself so it just get a smallest scope I was considering.

It may worth it.

src/lib.rs
jonhoo: My argument was more that all you need is the ability to wait forever. If the user wants to stop waiting before the action succeeds, they can just wrap the future in `tokio::time::timeout` :)

zhiburt:
I consider it's useful to have an abstractions `wait_for_not_visible`, `wait_for_present`, `wait_for_editable` etc. because it's not often clear how to implement them on it's own.

For example `element_is_editable` is pretty complicated to have implemented from end user.

```rust
     let is_displayed = match element.attr("style").await? {
            Some(style) if !style.contains("display: none;") => true,
            None => true,
             _ => false,
     };

      let is_enabled = match element.attr("disabled").await? {
           Some(..) => false,
           _ => true,
      };

      is_displayed && is_enabled
```

The timer logic could be shifted to `MutationObserver` if it'll work out.

zhiburt:
BTW: What do you think about adding `Element::style` method which would return a `std::collections::HashMap<String, String>` of styles since parsing `;`, `:` on it's own take a while :)

jonhoo:
I guess my argument is more that those methods should only be implemented in their "one shot" version, so `Element::is_editable` and `Element::is_present`. And then the user can just call those in a retry loop with a timer to get the `wait_for` behavior. It's not clear why we need to provide the looping version in the library itself.

Now, it might be useful to provide something like a `Wait` struct that users can use to easily get a retry loop, but I _don't_ think we should also implement tons of `wait_for_X` methods.

> By the way what do you think about adding `Element::style` method which would return a `std::collections::HashMap<String, String>` of styles since parsing `;`, `:` on it's own take a while :)

I'm not sure I follow? Are you talking about parsing the `style` tag? I really don't want to do that, because we probably don't want to get into trying to parse CSS strings, as they can get _very_ complicated (just splitting by `;` won't work). Besides, authors should probably be querying the current style using `.execute` and inspecting `.style`, rather than trying to parse out the `style` attribute.

PR #96, opened 2020-07-29
set_window_rect, set_window_rect, set_window_position parameters shift from i32 to u32

zhiburt:
Bind down the check `x >= 0` on type system.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -299,48 +299,18 @@ impl Client {
     }
 
     /// Sets the x, y, width, and height properties of the current window.
-    ///
-    /// All values must be `>= 0` or you will get a `CmdError::InvalidArgument`.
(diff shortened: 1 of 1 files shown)


jonhoo:
This is a good change, though it unfortunately is also backwards incompatible, so it might take a little while to land.

It also looks like CI is failing?

zhiburt:
There might have been some test which uses it but I forget to update them so it may be a cause.
I'll check this up.

zhiburt:
I've also checked the usage of functions which affected by this PR in test and they are fine.

jonhoo:
I merged #97, so maybe merge master into this now?

jonhoo:
Let's do it!

jonhoo:
Published in 0.14 :tada:

PR #99, opened 2020-08-05
Add active_element method

### LINKED ISSUE #98: Support getting the current active element?
Thanks for this library! I may be missing something, but is there a way to get the current active element for a session (via `/session/${session_id}/element/active`)? If not would you be willing to review a PR adding it?

travisbrown:
Resolves #98 via the WebDriver `GetActiveElement` command.

This method is useful for example when interacting with a site that provides keyboard shortcuts.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -501,6 +501,16 @@ impl Client {
         }
     }
 
+    /// Get the active element for this session.
+    pub async fn active_element(&mut self) -> Result<Element, error::CmdError> {
+        let res = self.issue(WebDriverCommand::GetActiveElement).await?;
+        let e = self.parse_lookup(res)?;
+        Ok(Element {
+            client: self.clone(),
+            element: e,
+        })
+    }
+
     /// Retrieve the currently active URL for this session.
     pub async fn current_url(&mut self) -> Result<url::Url, error::CmdError> {
         self.current_url_().await
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -502,6 +502,13 @@ impl Client {
     }
 
     /// Get the active element for this session.
+    ///
+    /// The "active" element is the `Element` within the DOM that currently has focus. This will
+    /// often be an `<input>` or `<textarea>` element that currently has the text selection, or
+    /// another input element such as a checkbox or radio button. Which elements are focusable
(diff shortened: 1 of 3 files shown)


jonhoo:
This seems like a great addition!

Two observations:
 - My read of the [active element spec](https://html.spec.whatwg.org/#dom-document-activeelement) and the [active element endpoint](https://www.w3.org/TR/webdriver1/#dfn-get-active-element) suggest that this might return `NoSuchElement`. I wonder — should we make the method return `Result<Option<Element>, ..>` for that case? And then a test for it as well?
 - I think we could probably provide a bit more helpful documentation for this method. The MDN has some [good text](https://developer.mozilla.org/en-US/docs/Web/API/DocumentOrShadowRoot/activeElement) that we could probably incorporate quite nicely!

travisbrown:
@jonhoo I'm happy to make that change, but it seems a little strange to me for `active_element` to return a `Result<Option<Element>, …>` when e.g. `find` returns a `Result<Element, …>`.

In the case of `find` the `NoSuchElement` failure isn't really exceptional, since any invalid selector will return it. For `active_element` my understanding is that most browsers will return the body if no element has focus, so it seems like it makes even more sense for the `NoSuchElement` possibility to be represented at the `Result` level (like it is for `find`) rather than an `Option` layer. (Or maybe I'm misunderstanding the decision not to make `find` return a `Result<Option…>>`?)

Definitely agree on documentation—will update that now.

travisbrown:
(Also it looks like the CI failure is unrelated, some issue with the geckodriver download.)

jonhoo:
> it seems a little strange to me for `active_element` to return a `Result<Option<Element>, …>` when e.g. `find` returns a `Result<Element, …>`.

Yeah, on second though, I think I agree with you. Let's keep it just `Result`. Happy to merge once the documentation update lands!

travisbrown:
Sounds good! I just pushed some additional documentation, paraphrased from MDN. I'd be happy to make further additions / revision.

jonhoo:
No, this looks great, thank you!

jonhoo:
Published in 0.14 :tada:

travisbrown:
Thanks!

Issue #57, opened 2019-09-14
Add a way to reconnect to an existing session


Follpvosten:
I don't think this is implemented yet, and I also don't really need it yet, but it would be nice to have; in my project, I may want to re-use the same session if the long-running process controlling it is restarted at some point.

At the moment, if I persist a session with `Client::persist` and then try to reconnect to the session (using `Client::with_capabilities` or `Client::new`), I get a `SessionNotCreated(WebDriverError { error: SessionNotCreated, message: "Session is already started", stack: "", delete_session: false })`.

I suppose the way to handle this would be adding a method like `Client::with_existing_session(webdriver: &str, session_id: &str)` and a corresponding `Session::from_id(webdriver: &str, id: &str)` or similar, which tries to connect to the given session using its ID (no idea how that's gonna work internally right now, but that would likely be the API). Then we can store the session ID and reconnect later.

I'll look into implementing this when I can find the time :)

jonhoo:
Awesome -- that sounds like a cool feature!

travisbrown:
@Follpvosten Have you looked into this? I implemented a very basic version today since I needed it myself, but it only does exactly what I need.

travisbrown:
@jonhoo Mind if I open a WIP PR with a simple version of this for the sake of discussion?

jonhoo:
Not at all, please do!

PR #100, opened 2020-08-06
Support reconnecting to an existing session

travisbrown:
This is a sketch of a minimal fix for #57 that just copies what `with_capabilities` is doing but without issuing the `NewSession` command.

I've confirmed that this implementation does what I need right now (allows me to have a CLI that doesn't have to log in to a site on every invocation, for example), but it should probably also do things like try to read the value of `is_legacy` off the WebDriver `status` for the session. It also needs tests, etc.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -242,6 +242,15 @@ impl Client {
         Session::with_capabilities(webdriver, cap).await
     }
 
+    /// Create a `Client` associated with an existing WebDriver session on the server at the given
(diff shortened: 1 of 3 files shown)


jonhoo:
Nice! I wonder if we could easily share some of this code between `new` and `with_session`. In theory I believe that should be doable.

JosiahBull:
Is this still being worked on?

PR #101, opened 2020-08-09
Update basic_setup example to use async await syntax

yerke:
Update basic_setup example to use async await syntax

Background: Today I tried to figure out how to use fantoccini, and opened basic_setup example first. I was very confused to find old futures syntax. Hopefully after these changes, newcomers will not have this problem.


yerke:
I am not sure I completely understand the difference between `wait_for_find()` and `find()`, but other examples in README and async_await use `find()` rather then `wait_for_find()`. So I switched the code to use `find()`.

jonhoo:
Oh, good catch! Have you also tried running it to see that it does the right thing?

yerke:
Of course!

Issue #86, opened 2020-05-09
Why do most methods use a mutable reference to Self?


ipkiss42:
Methods such as `Client.find()` or `Element.attr()` currrently take `&'_ mut self` as first parameter.

This is unexpected, and doesn't seem necessary. For example, removing the `mut` from both `Client.find()` and `Client.issue()` compiles fine.

Having a mutable self prevents from sharing the reference, but sharing it is needed in various situations, such as using it in the closure of a stream:

```rust
use futures::stream::{self, Stream, StreamExt};
use fantoccini::Client;

async fn do_stuff_with_client(client: &Client) -> u8 {
    // Some code here
    0
}

fn create_stream(client: &Client) -> impl Stream<Item = u8> + '_ {
    stream::iter(0..10).then(move |i| do_stuff_with_client(client))
}

async fn main_async() {
    let client = Client::new("http://localhost:4444").await.unwrap();
    let mut stream = Box::pin(create_stream(&client));
    while let Some(value) = stream.next().await {
        println!("{}", &value);
    }
}
```

The above sample code does not compile anymore if the client reference is mutable.

So, why have mutable references all over your API? Would you consider removing them?

jonhoo:
`Client` is `Clone` pretty much exactly so that you can clone the client handle if you wish. I don't want to make the methods take `&self` though, because, for example, it would mean that we are _required_ to use an unbounded channel here:
https://github.com/jonhoo/fantoccini/blob/93e849a8ca1b1eb1cb36b398eb425f2a23efb43c/src/session.rs#L22

Which I think we may want to move around from. It's also useful for the method to take `&mut self` just to remind the caller that they are mutating a single browser session. Arguably, `Client` shouldn't even be `Clone`, since the underlying webdriver connection handles requests one-at-a-time. But I decided that was too restrictive in practice.

ipkiss42:
Thanks, having `Clone` is a nice workaround to get unblocked.

From an API perspective it seems inconsistent though: either it is unsafe to have multiple mutating clients, and then `Clone` should also be forbidden (which is very restrictive, as you point out), or it is not, and the API should reflect it without the need to clone.

> I don't want to make the methods take `&self` though, because, for example, it would mean that we are _required_ to use an unbounded channel here:
> 
> https://github.com/jonhoo/fantoccini/blob/93e849a8ca1b1eb1cb36b398eb425f2a23efb43c/src/session.rs#L22

[...]

PR #104, opened 2020-09-20
lib: remove unnecessary mut

ernestas-poskus:
Change-Id: Idd2d253ec70c216b054bb166715bd2790eb83dd7

Sef mutabilit is unnecessary, removing it will allow to Send Client between threads.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -120,7 +120,7 @@
 //! [operators]: https://developer.mozilla.org/en-US/docs/Web/CSS/Attribute_selectors
 //! [WebDriver compatible]: https://github.com/Fyrd/caniuse/issues/2757#issuecomment-304529217
 //! [`geckodriver`]: https://github.com/mozilla/geckodriver
-#![deny(missing_docs)]
+#![deny(missing_docs, unused_mut)]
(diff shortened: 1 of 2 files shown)


jonhoo:
Hmm, I'm a little unsure about this. You may find the discussion over in https://github.com/jonhoo/fantoccini/issues/86 interesting. It's not entirely clear to me that it is better to provide a `&self` API over `&mut + Clone`.

ernestas-poskus:
>  which is a little unfortunate given that internally we already have a nice way of sharing access (the channel).

That was the reason why I have made this PR

> What really makes this complicated is that while these operations are logically concurrent, they still share a single connection to the webdriver. That's where all these concerns around &mut are coming from

Underlying channel ensures there's is a queue of tasks, therefore I find Clone inefficient in this case.

jonhoo:
The specific concern here is that it's not clear we'll want to continue to use a channel whose sender only takes `&self`. In which case we'd have to undo this change. I'd rather keep it the way it is, given that the `Clone` is very cheap compared to making HTTP requests anyway.

jonhoo:
These have now been changed in other ways.

PR #105, opened 2020-09-20
add with raw client request

ernestas-poskus:
Adding function that accepts raw hyper::Request 

cc @jonhoo

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -141,7 +141,7 @@ macro_rules! via_json {
     }};
 }
 
-pub use hyper::Method;
+pub use hyper::{self, Method};
 
 /// Error types.
 pub mod error;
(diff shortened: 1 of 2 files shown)


jonhoo:
Interesting — what is your use-case for this function? As it stands, it won't correctly include any cookies, or anything like that. To include those, the logic that `with_raw_client_for` includes would need to be invoked. I wonder if what you really want here is to write a `with_raw_client_for_request` which is the existing `with_raw_client_for` except that it modifies (and gets URL + method from) a passed-in `hyper::Request`. Then, you could change `with_raw_client_for` to just call `with_raw_client_for_request` with an empty `Request`.

I also feel like these methods should probably take request builders, not requests, as otherwise they won't be able to change headers for example. Maybe the "right" way to do this is to instead add a method `with_raw_client_for_request` that also takes a `F: FnOnce(hyper::RequestBuilder) -> hyper::Request` to be called after all the cookie stuff has been set up?

ernestas-poskus:
> what is your use-case for this function?

complete control over Request information 

> As it stands, it won't correctly include any cookies, or anything like that.

this is expected 

> I also feel like these methods should probably take request builders

IMO "raw" in function signature expects the user to understand what one is doing

jonhoo:
In this case there's little more to the request than just creating your own `hyper::Request` to `current_url()`, so why have a dedicated method for it?

PR #106, opened 2020-10-02
Add support for fetching css values of an element

awikman:
Implementing the `GET  /session/{session id}/element/{element id}/css/{property name}` method.

I needed this to fetch the background-color of an element, so an example use case would be something like:
```
  let sel_body = r#"body"#;
  let mut el_body = client.find(Locator::Css(sel_body)).await?;
  let body_background_color = el_body.css("background-color").await?;
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -987,6 +987,16 @@ impl Element {
             .collect())
     }
 
+    /// Look up a CSS value for this element by name (color, width, etc)
+    pub async fn css(&mut self, prop: &str) -> Result<Option<String>, error::CmdError> {
+        let cmd = WebDriverCommand::GetCSSValue(self.element.clone(), prop.to_string());
+        match self.client.issue(cmd).await? {
+            Json::String(v) => Ok(Some(v)),
+            Json::Null => Ok(None),
+            v => Err(error::CmdError::NotW3C(v)),
+        }
+    }
+
     /// Simulate the user clicking on this element.
     ///
     /// Note that since this *may* result in navigation, we give up the handle to the element.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -988,11 +988,12 @@ impl Element {
     }
 
     /// Look up a CSS value for this element by name (color, width, etc)
+    ///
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -987,6 +987,16 @@ impl Element {
             .collect())
     }
 
+    /// Look up a CSS value for this element by name (color, width, etc)
+    pub async fn css(&mut self, prop: &str) -> Result<Option<String>, error::CmdError> {
+        let cmd = WebDriverCommand::GetCSSValue(self.element.clone(), prop.to_string());
+        match self.client.issue(cmd).await? {
+            Json::String(v) => Ok(Some(v)),
+            Json::Null => Ok(None),
```
jonhoo, line 995: What exactly does `None` mean here? Does it mean that the element did not exist, that the attribute was not set, or something else? Would be good to add that to the method documentation!

jonhoo:
This looks good with just a little more documentation! If you could also add a simple test, that'd be great :)

src/lib.rs
awikman: Looks like the css command always returns a string, even if the property is not defined. Must have been some issue with my testing previously.... 14a90c8

awikman:
I'll try to add more as soon as I have time.
Looks like the tests are currently failing. Is there PR/Issue about that already?

src/lib.rs
```diff
@@ -987,6 +987,17 @@ impl Element {
             .collect())
     }
 
+    /// Look up a CSS value for this element by name (color, width, etc)
+    ///
+    /// See <https://www.w3.org/TR/webdriver/#dfn-get-element-css-value>.
+    pub async fn css(&mut self, prop: &str) -> Result<Option<String>, error::CmdError> {
```
jonhoo, line 993: Is there still a reason for the method to return `Option<String>`, or should that just be `String` now?

As for documentation, I'd like to see it explicitly document what the caller should expect to get back if the property is malformed or hasn't been set.

jonhoo:
This is now present in [0.19.0-rc.3](https://docs.rs/fantoccini/0.19.0-rc.3/fantoccini/elements/struct.Element.html#method.css_value) and onwards :)

PR #107, opened 2020-10-07
Add execute_async cmd

nkconnor:
Endpoint implemented: https://github.com/jlipps/simple-wd-spec#execute-async-script

Todo, from my perspective:

- [x] Check that docs build and render correctly
- [x] See if `CmdError` needs a `Timeout` variant

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -86,6 +86,13 @@ pub enum CmdError {
     /// ["no such window"]: https://www.w3.org/TR/webdriver/#dfn-no-such-window
     NoSuchWindow(wderror::WebDriverError),
 
+    /// The allowed script execution time was exceeded. This variant lifts the
+    /// ["script timeout"] error variant from `Standard` to simplify checking for it in
+    /// user code
+    ///
+    /// ["script timeout"]: https://www.w3.org/TR/webdriver/#dfn-script-timeout-error
+    ScriptTimeout(wderror::WebDriverError),
+
     /// A bad URL was encountered during parsing.
     ///
     /// This normally happens if a link is clicked or the current URL is requested, but the URL in
@@ -154,6 +161,7 @@ impl Error for CmdError {
             CmdError::NotW3C(..) => "webdriver returned non-conforming response",
             CmdError::InvalidArgument(..) => "invalid argument provided",
             CmdError::ImageDecodeError(..) => "error decoding image",
+            CmdError::ScriptTimeout(..) => "script timeout",
         }
     }
 
@@ -167,7 +175,10 @@ impl Error for CmdError {
             CmdError::Lost(ref e) => Some(e),
             CmdError::Json(ref e) => Some(e),
             CmdError::ImageDecodeError(ref e) => Some(e),
-            CmdError::NotJson(_) | CmdError::NotW3C(_) | CmdError::InvalidArgument(..) => None,
+            CmdError::NotJson(_)
+            | CmdError::NotW3C(_)
+            | CmdError::InvalidArgument(..)
+            | CmdError::ScriptTimeout(_) => None,
         }
     }
 }
@@ -187,6 +198,7 @@ impl fmt::Display for CmdError {
             CmdError::Json(ref e) => write!(f, "{}", e),
             CmdError::NotW3C(ref e) => write!(f, "{:?}", e),
             CmdError::ImageDecodeError(ref e) => write!(f, "{:?}", e),
+            CmdError::ScriptTimeout(ref e) => write!(f, "{:?}", e),
             CmdError::InvalidArgument(ref arg, ref msg) => {
                 write!(f, "Invalid argument `{}`: {}", arg, msg)
             }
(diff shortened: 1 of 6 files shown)


nkconnor:
Note to self: 

```
NoSuchWindow(WebDriverError)

The requested window does not exist.

This variant lifts the "no such window" error variant from Standard to simplify checking for it in user code.
```

I didn't see this code when adding the `ScriptTimeout` variant :shrug:

src/lib.rs
```diff
@@ -514,6 +514,37 @@ impl Client {
...
+    /// with the provided value. For example,
+    ///
+    /// ```javascript
+    /// const done = arguments.pop();
+    ///
+    /// setInterval(() => {
+    ///     // ... do stuff
+    ///     done(valueToReturn);
+    /// }, 5000);
+    /// ```
```
jonhoo, line 533: I don't quite follow this example — could you include the Rust part of the example to help illustrate the callback mechanism for the last argument?

src/error.rs
```diff
@@ -86,6 +86,13 @@ pub enum CmdError {
     /// ["no such window"]: https://www.w3.org/TR/webdriver/#dfn-no-such-window
     NoSuchWindow(wderror::WebDriverError),
 
+    /// The allowed script execution time was exceeded. This variant lifts the
+    /// ["script timeout"] error variant from `Standard` to simplify checking for it in
+    /// user code
+    ///
+    /// ["script timeout"]: https://www.w3.org/TR/webdriver/#dfn-script-timeout-error
+    ScriptTimeout(wderror::WebDriverError),
```
jonhoo, line 94: Ah, this is pretty unfortunate. Since the public error type here doesn't have `#[non_exhaustive]`, adding a new variant is a breaking change, and we'd have to release a new major version. Is this error variant commonly encountered?

nkconnor: Oops! Thanks for bringing that up. The default is 30 seconds, which, for a lot of use cases is plenty. I'm not sure what most people are using the project for but I'd guess this is somewhat common. One example is if you execute an async function and only call the callback on success ("programmer error"), and another would be say waiting for an API call to return. 

I actually have a follow-up commit that has `Client::set_script_timeout` because my use case takes so long.

I'm open to editing however you see fit. Or perhaps a new branch w/ this to anticipate a release some day later.

src/lib.rs
nkconnor: Added some changes here which I hope makes it clearer: https://github.com/jonhoo/fantoccini/pull/107/commits/b210632ba868dc867a4cec3b7af849a828910632

jonhoo: Ah, I think what was missing to make this click for me was that this function automatically makes the arguments available as the `arguments` variable in the JS code. I wonder if we can make this more explicit with some language along the lines of:
> The provided JavaScript has access to the provided arguments through the `arguments` array. The array also holds an additional element at the end that provides a completion callback for the asynchronous code. [..]

src/error.rs
jonhoo: Hmm... I wonder if the right thing to do here is to keep `ScriptTimeout` errors as nested under `Standard`. After all, that's sort of what `Standard` is there for. `NoSuchElement` and `NoSuchWindow` are just explicitly extracted out because they are _so_ common that it'd be annoying to go fishing them out of `Standard` each time.

src/lib.rs
nkconnor: Hmm, GitHub is acting oddly for me, still showing the prior docs in the review. How does this look?

```
    /// Execute the given async JavaScript script in the current browser session.
    ///
    /// The provided JavaScript has access to the provided `args` through the `arguments` array. The array also 
    /// holds an additional element at the end that provides a completion callback for the asynchronous code.
    ///
    /// Since `Element` implements `Serialize`, you can also provide serialized `Element`s as arguments, and they will
    /// correctly deserialize to DOM elements on the other side.
    ///
    /// # Examples
    ///
    /// Call a web API from the browser and retrieve the value asynchronously
    ///
    /// ```ignore
    /// const JS: &'static str = r#"
    ///     const [date, callback] = arguments;
    ///
    ///     fetch(`http://weather.api/${date}/hourly`)
    ///     // whenever the HTTP Request completes,
    ///     // send the value back to the Rust context
    ///     .then(data => {
    ///         callback(data.json())
    ///     })
    /// "#;
    ///
    /// let weather = client.execute_async(JS, vec![date]).await?;
    /// ```
```

src/error.rs
```diff
@@ -214,14 +214,12 @@ impl From<herror::Error> for CmdError {
...
     fn from(e: wderror::WebDriverError) -> Self {
-        if let wderror::WebDriverError {
-            error: wderror::ErrorStatus::NoSuchElement,
-            ..
-        } = e
-        {
-            CmdError::NoSuchElement(e)
-        } else {
-            CmdError::Standard(e)
+        match e {
```
jonhoo, line 217: Maybe undo this change entirely (make it be `if let` again) now that we're no longer looking for script timeout?

src/lib.rs
jonhoo: Yeah, that looks much better. Thinking some more about it, I think we want to be _even_ more explicit:
> The provided JavaScript has access to `args` through the JavaScript variable `arguments`.

nkconnor: Updated again, thanks!

jonhoo (approved):

jonhoo:
This looks excellent now, thank you! Once CI passes, I'll merge and publish :)
Thanks for sticking with all my comments!

jonhoo:
Could you merge `master` into this branch to get the CI fix I just pushed?

nkconnor:
Should be in now

PR #109, opened 2020-11-04
Add select_by_label function

zhiburt:
It seems like it works but I am not sure if it is :)

___

To note: I've noticed while I was going though a selenium lib that they 'sometimes'  escape quotes in some function (like this). If it is an important thing I would imagine that it would affect not only this function so I just note it here.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1133,6 +1133,12 @@ impl Element {
         self.find(Locator::Css(&locator)).await?.click().await
     }
 
+    /// Find and click an `<option>` element by its visible text/label.
+    pub async fn select_by_label(mut self, label: &str) -> Result<Client, error::CmdError> {
+        let locator = format!(r".//option[normalize-space(.)='{}']", label);
+        self.find(Locator::XPath(&locator)).await?.click().await
+    }
+
     /// Switches to the frame contained within the element.
     pub async fn enter_frame(self) -> Result<Client, error::CmdError> {
         let Self {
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -1133,6 +1133,12 @@ impl Element {
         self.find(Locator::Css(&locator)).await?.click().await
     }
 
+    /// Find and click an `<option>` element by its visible text/label.
+    pub async fn select_by_label(mut self, label: &str) -> Result<Client, error::CmdError> {
+        let locator = format!(r".//option[normalize-space(.)='{}']", label);
```
jonhoo, line 1138: What does the `normalize-space` do here?
As for quoting, that's a tricky one. I think trying to do quoting ourselves is just going to lead to problems, so I'd probably mention in the documentation that it's up to the caller to escape `'` characters :)

jonhoo:
Sorry for the slow reply!

src/lib.rs
zhiburt: As I can state `normalize-space` was used to handle invisible from IDEAs perspective sings like `&amp` etc.

Overall looking for something by the label is a tricky one on its own. So I think you're right that we should put the responsibility for the proper lookup string on the caller.

I'll update the PR in the nearest (I hope)

jonhoo (approved):

jonhoo:
Looks great, thank you!

PR #110, opened 2020-11-04
Add a select_by_id command

zhiburt:
As you may see it's considered to be a 4 functions which select element which could provide a high level function like `select` which would expect `by` parameter or something similar.

```rust
element.select(SelectBy::Value("".to_string));
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1133,6 +1133,15 @@ impl Element {
         self.find(Locator::Css(&locator)).await?.click().await
     }
 
+    /// Find and click an `<option>` child element by its id.
+    ///
+    /// This method clicks the first `<option>` element that has an appropriate `id`.
+    /// If the element wasn't found a erorr will be issued.
+    pub async fn select_by_id(mut self, id: &str) -> Result<Client, error::CmdError> {
+        let locator = format!(r#"option[id='{}']"#, id);
+        self.find(Locator::Css(&locator)).await?.click().await
+    }
+
     /// Switches to the frame contained within the element.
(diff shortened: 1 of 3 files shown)


jonhoo:
So, my thinking here for adding the methods that current exist was that it's actually kind of hard to write the selector yourself. Something like `#optionid` on the other hand is very straightforward for the client to write, which is why there isn't a method for it. And I don't think it's really worthwhile adding one?

zhiburt:
Do you say it's no worth it to add a `select_by_id`?

The thing why it may be useful is because there's a bunch of people out there who have no idea that select actually works by clicking. So it may save them time (kinda)

If you still consider it's not worth to increase a binary size I can see what are you standing for :smile:

jonhoo:
I'm more thinking that I don't think we want to add a bunch of "trivial" methods to the API, since it makes the API more intimidating to new users. Instead, maybe it'd be good to have a `select_by_css`?

zhiburt:
Maybe `select_by` in this case is better so user could use `xpath` for example?

jonhoo:
Sure, that seems fine too :)

jonhoo (approved):

jonhoo:
@zhiburt This looks good now, though it looks like the test suite doesn't pass for your documentation tests (`cargo test --doc`):
https://travis-ci.com/github/jonhoo/fantoccini/jobs/441475467#L1222

zhiburt:
Well it was a while :)
Coincidentally got back here :)

Now I am looking at `select_by` function and I am thinking is it useful to be exposed as `pub`lic function at all.

What do you think?

PR #115, opened 2020-12-26
Move to rustls to stop linking to openssl

x0f5c3:
This is probably a personal preference in a way, but I have problems with other rust libraries and software that link to openssl and it can be hard to compile it on musl or without installing openssl libraries on the system so I think this will make it easier to deploy an application on alpine containers or on clean systems without openssl. This change switches hyper to a rustls connector and in my testing it passes all tests and compiles on musl.


jonhoo:
I'm skeptical of making this change specifically because it also means that it's now difficult for people who *do* want to use OpenSSL. I think the right path forward is actually probably for `Session` to be generic over the `hyper` connector. Specifically, for it to be `Session<C>` where `C: HttpConnector`. And then we could pretty easily make the HTTPS connector and on-by-default feature or something. What do you think?

x0f5c3:
That's a great idea, only problem I had to resort to couple of cfg attributes and 2 functions for constructing the connector because to add trait bounds would require to provide associated types for those traits like Service or Connect, other option is to add an enum or a struct to hold the connector and just add a use for it hidden behind a cfg attribute. I set it up so in case all-features are chosen rustls is the default for now.

jonhoo:
Ah, unfortunately that won't work since Rust features must be additive — it must always be possible to add all features, and adding a feature cannot make code stop compiling.

What I was imagining was that we make `Session` generic over `C`m and then store a `hyper::Client<C>` instead. And then we provide a `new` that takes a `C`, as well as a `new_hyper_tls` under a `hyper-tls` feature, and a `new_rusttls` under a `rusttls` feature. It'll be a backwards-incompatible change I think, but a worthwhile one.

x0f5c3:
I added a trait alias to make it more concise and I wanted to ask how I should approach Client::new() so for now it only makes the client with rustls, but the session should be ready for both

src/session.rs
```diff
@@ -173,22 +173,33 @@ impl Ongoing {
         Poll::Ready(rt)
     }
 }
+pub(crate) trait Connector:
```
jonhoo, line 176: Let's stick to just giving the bounds directly. Intermediate traits tend to cause pain with type inference. One day we'll have proper trait aliases, but for now let's repeat the bound :)

src/session.rs
```diff
@@ -173,22 +173,33 @@ impl Ongoing {
...
+    Sized + Sync + Clone + hyper::client::connect::Connect + Unpin + Send + 'static
```
jonhoo, line 177: I don't think we need `Sized` or `Unpin` here. We _probably_ don't need `Send + Sync + 'static` either?

src/lib.rs
```diff
@@ -239,7 +244,10 @@ impl Client {
         webdriver: &str,
         cap: webdriver::capabilities::Capabilities,
     ) -> Result<Self, error::NewSessionError> {
-        Session::with_capabilities(webdriver, cap).await
+        Session::<hyper_rustls::HttpsConnector<hyper::client::HttpConnector>>::new_rustls(
```
jonhoo, line 247: I think we'll probably want `Client` to be generic over `C` as well.

src/session.rs
```diff
@@ -327,19 +344,34 @@ impl Session {
...
+        cap: webdriver::capabilities::Capabilities,
+    ) -> Result<Client, error::NewSessionError> {
+        let connector = hyper_rustls::HttpsConnector::new();
+        Self::with_capabilities(webdriver, cap, connector).await
+    }
 
     pub(crate) async fn with_capabilities(
         webdriver: &str,
         mut cap: webdriver::capabilities::Capabilities,
+        connector: impl Connector,
```
jonhoo, line 366:
```suggestion
        connector: C,
```

src/session.rs
```diff
@@ -327,19 +344,34 @@ impl Session {
...
+        let connector = hyper_tls::HttpsConnector::new();
+        Self::with_capabilities(webdriver, cap, connector).await
+    }
+    pub(crate) async fn new_rustls(
+        webdriver: &str,
+        cap: webdriver::capabilities::Capabilities,
+    ) -> Result<Client, error::NewSessionError> {
+        let connector = hyper_rustls::HttpsConnector::new();
+        Self::with_capabilities(webdriver, cap, connector).await
+    }
```
jonhoo, line 361: These `impl`s probably need to be in their own `impl` blocks for the specific `Client<C>` that they each produce.

src/session.rs
x0f5c3: Yeah, probably :), Shame there are no trait aliases yet. I'll change it to normal bounds

src/session.rs
x0f5c3: From what I understand we need Send + Sync + 'static to satisfy bounds on hyper::client::Client implementation of Service and Unpin to impl Future and also for the tokio::spawn call in creating the client, maybe I'm going at it wrong but it works for me only with these bounds.

src/lib.rs
x0f5c3: I think that could create more problems since Element and Form both have Client in them and maybe we could do it like this:

```rust
pub(crate) async fn with_capabilities<T>(
        webdriver: &str,
        mut cap: webdriver::capabilities::Capabilities,
        connector: T,
    ) -> Result<Client, error::NewSessionError>
    where
        T: Clone + hyper::client::connect::Connect + Sync + Send + 'static + Unpin,
```
And maybe indicate with an enum passed to the call to client new to indicate which version we want if both features are enabled and internally use a type parameter in creating the Session
That would also eliminate the need on the user's part to use hyper themselves

src/session.rs
jonhoo: Hmm, you may need them for certain `impl` blocks, but I don't think you need all of them everywhere?

src/lib.rs
jonhoo: Actually, maybe the real way to go about this is to store a `Box<dyn Connect + ...>` inside `Session` rather than making it generic. Dynamic dispatch should be fine here, and would make the interface much nicer!

x0f5c3:
This is what I came up with, the problem with dynamic dispatch is the associated type _Svc in the Connect trait, I'm probably going to have to find a better way to let the user choose between OpenSSL and RustTLS but I couldn't find a way to make the traits and generics differently

x0f5c3:
I also had an idea taken from a project I'm working on using this crate to simplify capabilities a little bit, maybe I could add it to this:

```rust
pub trait CapabilitiesExt {
    fn headless(self) -> Self;
}
impl CapabilitiesExt for webdriver::capabilities::Capabilities {
    fn headless(mut self) -> Self {
        let arg = serde_json::json!({"args": ["-headless"]});
        self.insert("moz:firefoxOptions".to_string(), arg);
        self
    }
}
```
Something along those lines

jonhoo:
Ah, unfortunately, this approach won't work either -- you can't have function signatures differ depending on a feature. And you're right, dynamic dispatch won't work either since `Connect` isn't object-safe. I think our only path forward here is to add a `<C>` generic type on all the types then. It's not pretty, but it's the only way we can really do this. And then we can just keep `new` and `with_capabilities` the way they are today, and also add two new functions: `with_connector` and `with_capabilities_and_connector`.

Your idea for an extension trait is a good one -- I think that's a great idea!

x0f5c3:
Yes, I think we need to add bounds on all the types. By the way, in Client methods do we need to have it on &mut self and not on &self? I changed almost all of them to &self and the firefox tests pass and I think if we can do it with immutable borrows that would simplify using the library since the user could just pass the reference back and forth.

x0f5c3:
I added a trait to construct the two httpsconnectors so Session::with_capabilities can run C::new() and I added with_connector and with_connector_and_capabilities to Client. For now the default Client::new() and Client::with_capabilities uses rustls but that can be changed.

jonhoo:
You should be able to use `C` by sticking a `PhantomData` in there!

x0f5c3:
I changed it to `PhantomData`, I think that the generics can complicate stuff for the user because of the need to write the types on every argument and return value that contain `Client`, `Element` or `Form` but this way is also more explicit and lets the user decide on the connector. The `NewConnector` trait allows for the Session to do C::new() but it can also be done by doing two seperate impl for `Session`, one for `hyper_rustls` and other `hyper_tls`

jonhoo:
I don't think we need to have `NewConnect` as a bound _everywhere_. It probably only needs to be there for the constructor. Everywhere else I think we only need `Connect`, no?

src/lib.rs
```diff
@@ -197,36 +215,95 @@ pub use crate::session::Client;
...
+    C: NewConnector + connect::Connect,
+{
+    /// Create a new [`Client`][crate::Client] associated with a new WebDriver session on the server at the given
     /// URL.
     ///
+    ///
+    ///
     /// Calls `with_capabilities` with an empty capabilities list.
     #[allow(clippy::new_ret_no_self)]
     pub async fn new(webdriver: &str) -> Result<Self, error::NewSessionError> {
```
jonhoo, line 252: I think what we should do is have `new` use `#[cfg(feature = "hyper-tls")]`, and specifically construct a `Client<hyper_tls::HttpsConnector>`. That way there's still an easy way for users to create a client that "just works" if they don't care too much about the underlying TLS stack, and you can remove all the `<...>` annotations you've had to add to the examples.

src/lib.rs
```diff
@@ -197,36 +215,95 @@ pub use crate::session::Client;
...
-    /// Create a new `Client` associated with a new WebDriver session on the server at the given
+impl<C> Client<C>
+where
+    C: NewConnector + connect::Connect,
+{
+    /// Create a new [`Client`][crate::Client] associated with a new WebDriver session on the server at the given
     /// URL.
     ///
+    ///
+    ///
```
jonhoo, line 249: Why these extra newlines?

src/session.rs
```diff
@@ -327,19 +358,129 @@ impl Session {
             }
         }
     }
+    pub(crate) async fn with_capabilities_and_connector(
```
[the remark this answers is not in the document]
x0f5c3: I can make with_capabilities have connector: Option<C> and just check if it's there or not and if not construct the connector, I don't know if there's a way to do it without the Option and without copy-pasting the entire thing.

```rust
 pub(crate) async fn with_capabilities(
        webdriver: &str,
        mut cap: webdriver::capabilities::Capabilities,
        connector: Option<C>,
    ) -> Result<Client<C>, error::NewSessionError> {
        // Where is the WebDriver server?
        let wdb = webdriver.parse::<url::Url>();
        let wdb = wdb.map_err(error::NewSessionError::BadWebdriverUrl)?;
        let connector = if let Some(conn) = connector {
            conn
        } else {
            C::new()
        };
```

src/lib.rs
x0f5c3: So just leave new only on feature `openssl-tls` or `rustls-tls` as a default and in the case that user does default-features = false then they have to use with_capabilities with type parameters?
Also we could just do type aliases for the two connectors, that way the user can just `use fantoccini::Rustls` or `use fantoccini::OpenSSL` and just use this as a type parameter without the need to add hyper and hyper_rustls or hyper_tls to their dependencies

x0f5c3:
> 
> 
> I don't think we need to have `NewConnect` as a bound _everywhere_. It probably only needs to be there for the constructor. Everywhere else I think we only need `Connect`, no?

The `NewConnector` trait acts also as a bound for `Clone + Send + Sync + Unpin + 'static` so we would then need to add those instead

x0f5c3:
Maybe this way, with type aliases in case of only one enabled features and pub use Client when both are enabled can simplify it for the end user

jonhoo:
Oh, please don't move all of `Client` into its own in this PR -- it makes it impossible to see what's actually changed (mostly because GitHub can't really show the diff). If I understand you correctly though, I don't think that'll work. Remember, features _must_ be additive, so if both features are enabled, the code must still compile, and any code that compiles against the crate with one or the other feature enabled must continue to compile.

I was proposing we _only_ expose `new` if the `hyper-tls` feature is used (_not_ if only the `rusttls` feature is given).

Two separate type aliases for `Client<hyper_tls::...>` and for `Client<rusttls::...>` might be a good idea though -- I like that!

I don't think we need `Clone + Send + Sync + Unpin + 'static` on every `impl` block. Probably only on a fairly small number of them. I'm proposing we only add the bounds we _need_ in the places we need them.

x0f5c3:
> 
> 
> Oh, please don't move all of `Client` into its own in this PR -- it makes it impossible to see what's actually changed (mostly because GitHub can't really show the diff). If I understand you correctly though, I don't think that'll work. Remember, features _must_ be additive, so if both features are enabled, the code must still compile, and any code that compiles against the crate with one or the other feature enabled must continue to compile.
> 
> I was proposing we _only_ expose `new` if the `hyper-tls` feature is used (_not_ if only the `rusttls` feature is given).
> 
> Two separate type aliases for `Client<hyper_tls::...>` and for `Client<rusttls::...>` might be a good idea though -- I like that!
> 
> I don't think we need `Clone + Send + Sync + Unpin + 'static` on every `impl` block. Probably only on a fairly small number of them. I'm proposing we only add the bounds we _need_ in the places we need them.

You're right, I'm going to move back the code from client.rs to lib.rs, the type aliases seem like the reasonable way to proceed, if only one of the features is enabled then the user just has to do `Client::new()` and no need for type parameters. I'll try to cut down on the bounds but most of the impl blocks require them, at least `Clone + Send + Sync` are needed on all impl blocks due to these ones being required for all connectors for `hyper`, maybe in some places we could get away without `'static` and maybe without `Unpin` but both of them are needed anywhere there's `tokio::spawn` and in Future impls. Right now the code compiles and runs tests succesfully on one and on all features, it compiles with no features enabled but if the user does it this way then they need to just impl NewConnector on their `HttpsConnector` and it should work.

x0f5c3:
I removed trait bounds from `NewConnector` trait but I still needed `Unpin + 'static` on `impl Future` and `impl Session` for `Session` so in turn I also needed them on `impl Client`, which caused them to be needed on `impl Element` and `Form` and so on. So I think it's only a matter of preference whether we keep them on the `NewConnector` trait or on `impl`

jonhoo:
> You're right, I'm going to move back the code from client.rs to lib.rs, the type aliases seem like the reasonable way to proceed, if only one of the features is enabled then the user just has to do `Client::new()` and no need for type parameters.

Ah, unfortunately, this is still not okay. It might help to see an example of why not. Imagine that we have two crates, `a` and `b`, that both depend on `fantoccini`:
```toml
name = "a"
[dependencies]
fantoccini = { version = "*", features = ["rustls-tls"] }
```
```toml
name = "b"
[dependencies]
fantoccini = { version = "*", features = ["openssl-tls"] }
```
Since they both just use one feature, they both contain code that just uses `let c = Client::new()`.

Now imagine that a crate `c` comes along that depends on both `a` **and** `b`. When cargo now compiles `fantoccini`, it will compile it with _both_ features enabled (it takes the union of the features required by all users of the crate). When that happens, neither `a` nor `b` will compile, since `Client::new` is no longer valid.

What I was proposing instead was to have two _different_ type aliases (or just constructor methods), one for `openssl-tls`, and one for `rustls-tls`.

> Right now the code compiles and runs tests succesfully on one and on all features, it compiles with no features enabled but if the user does it this way then they need to just `impl NewConnector` on their `HttpsConnector` and it should work.

I'd like to avoid users ever having to implement `NewConnector`. Arguably, the trait should be sealed, since it's really more of an implementation detail.

> So I think it's only a matter of preference whether we keep them on the `NewConnector` trait or on `impl`

I don't quite follow this — `NewConnector` has a method that lets you create a _new_ connector. That should only be needed when you first create a client/session, and never again. All other `impl` blocks should only require `Connect`, plus some other traits here and there, they should not need access to a `new` method? And given Rust's (current) lack for a trait alias feature, I'd like to just repeat the bounds for now. They can be simplified later through something like implied bounds or true type aliases, and I'd like to avoid adding a new trait just for that forwarding, since it inevitably causes issues with type inference in my experience.

x0f5c3:
> What I was proposing instead was to have two _different_ type aliases (or just constructor methods), one for `openssl-tls`, and one for `rustls-tls`.

Ok, I understand, I was going at it from a direction of wanting to simplify the `Client` for the end-user but yeah from the example given here that would bring problems if two crates depended on different connectors.

> I'd like to avoid users ever having to implement `NewConnector`. Arguably, the trait should be sealed, since it's really more of an implementation detail.


I was just thinking that maybe users could plug their own `HttpsConnectors` like for BoringSSL etc. but it can be sealed and users just use those two.

> I don't quite follow this — `NewConnector` has a method that lets you create a _new_ connector. That should only be needed when you first create a client/session, and never again. All other `impl` blocks should only require `Connect`, plus some other traits here and there, they should not need access to a `new` method? And given Rust's (current) lack for a trait alias feature, I'd like to just repeat the bounds for now. They can be simplified later through something like implied bounds or true type aliases, and I'd like to avoid adding a new trait just for that forwarding, since it inevitably causes issues with type inference in my experience.

It should work with `NewConnector` only as a bound in `impl<C> Session<C>` and all trait bounds on all impl blocks

x0f5c3:
I added type aliases for both variants of Client and sealed the trait and only used it for creating the Client and Session

jonhoo:
> > What I was proposing instead was to have two _different_ type aliases (or just constructor methods), one for `openssl-tls`, and one for `rustls-tls`.
> 
> Ok, I understand, I was going at it from a direction of wanting to simplify the `Client` for the end-user but yeah from the example given here that would bring problems if two crates depended on different connectors.

Yeah, it's an unfortunate trade-off, but given that this is how cargo treats features (and for good reason), we sort of need to do it that way!

> > I'd like to avoid users ever having to implement `NewConnector`. Arguably, the trait should be sealed, since it's really more of an implementation detail.
> 
> I was just thinking that maybe users could plug their own `HttpsConnectors` like for BoringSSL etc. but it can be sealed and users just use those two.

I'm imagining that users of other connectors would simply use `with_connector`, so I don't think this precludes any other use-cases.

> 
> > I don't quite follow this — `NewConnector` has a method that lets you create a _new_ connector. That should only be needed when you first create a client/session, and never again. All other `impl` blocks should only require `Connect`, plus some other traits here and there, they should not need access to a `new` method? And given Rust's (current) lack for a trait alias feature, I'd like to just repeat the bounds for now. They can be simplified later through something like implied bounds or true type aliases, and I'd like to avoid adding a new trait just for that forwarding, since it inevitably causes issues with type inference in my experience.
> 
> It should work with `NewConnector` only as a bound in `impl<C> Session<C>` and all trait bounds on all impl blocks

Shouldn't _only_ the very first constructor have the `NewConnector` bound? So basically it should be a requirement of `Client::new`, `Client::with_capabilities`, and `Session::with_capabilities`, but no other methods. In particular, if someone constructs a new `Client` using `Client::with_connector` with a `C` that does _not_ implement `NewConnector`, they should still then be able to call all other methods.

x0f5c3:
You're right, I forgot that we could just leave with_connector methods in impls with normal trait bounds without NewConnector and then any would work, sorry about that.

src/session.rs
```diff
@@ -296,14 +296,15 @@ impl<C> Session<C>
 where
     C: connect::Connect + Unpin + 'static + Clone + Send + Sync + NewConnector,
 {
-    pub(crate) async fn with_capabilities_and_connector(
+
+    pub(crate) async fn with_capabilities(
```
jonhoo, line 300: This should just call `with_capabilities_and_connector` with `C::new`.

src/lib.rs
```diff
@@ -123,6 +154,34 @@
 #![deny(missing_docs)]
 #![warn(missing_debug_implementations, rust_2018_idioms)]
 
+
+
+pub use hyper::Method;
```
jonhoo, line 159: Why move this and the other `use` and `mod` statements from where they were before?

src/lib.rs
```diff
@@ -29,12 +34,23 @@
 //! ```no_run
 //! # extern crate tokio;
 //! # extern crate fantoccini;
-//! use fantoccini::{Client, Locator};
+//! # #[cfg(feature = "rustls-tls")]
```
jonhoo, line 37: This example now won't compile if no feature are enabled (which is also a valid way to compile the crate). I suggest that in these examples we just use the `openssl-tls` version, put
```rust
#![cfg(feature = "openssl-tls")]
```
at the top of the example (so it'll be ignored when the feature isn't enabled), and the explain in the documentation to see a later section of the docs for details on how to use a different TLS implementation.

src/lib.rs
```diff
@@ -123,6 +154,34 @@
...
+
+/// Type alias for a Client with a Rustls connector
+///
+#[cfg(feature = "rustls-tls")]
+pub type RustlsClient =  session::Client<hyper_rustls::HttpsConnector<hyper::client::HttpConnector>>;
+
+/// Type alias for a Client with an OpenSSL connector
+///
+#[cfg(feature = "openssl-tls")]
+pub type OpenSslClient = session::Client<hyper_tls::HttpsConnector<hyper::client::HttpConnector>>;
```
jonhoo, line 178: I wonder, what if instead of having these aliases, we instead just provide the following `impl` blocks:
```rust
#[cfg(feature = "rustls-tls")]
impl Client<hyper_rustls::HttpsConnector<hyper::client::HttpConnector>> {
    pub fn new_rusttls() -> Self {
        Client::with_connector(hyper_rustls::HttpsConnector<hyper::client::HttpConnector>::new())
    }
}

#[cfg(feature = "openssl-tls")]
impl Client<hyper_tls::HttpsConnector<hyper::client::HttpConnector>> {
    pub fn new() -> Self {
        Client::with_connector(hyper_tls::HttpsConnector<hyper::client::HttpConnector>::new())
    }
}
```
That way we can also get rid of the `NewConnector` trait entirely.

src/lib.rs
```diff
@@ -193,40 +248,54 @@ impl<'a> From<Locator<'a>> for webdriver::command::LocatorParameters {
...
+    where
+        C: connect::Connect + Send + Sync + Clone,
+{
     #[serde(skip_serializing)]
-    client: Client,
+    client: session::Client<C>,
     #[serde(flatten)]
     element: webdriver::common::WebElement,
+    #[serde(skip_serializing)]
+    _marker: marker::PhantomData<C>,
```
jonhoo, line 263: I don't think you need this any more now?

src/lib.rs
```diff
@@ -193,40 +248,54 @@ impl<'a> From<Locator<'a>> for webdriver::command::LocatorParameters {
...
 #[derive(Clone, Debug)]
-pub struct Form {
-    client: Client,
+pub struct Form<C>
+    where
+        C: connect::Connect + Send + Sync + Clone,
+{
+    client: session::Client<C>,
     form: webdriver::common::WebElement,
+    _marker: marker::PhantomData<C>,
```
jonhoo, line 274: I don't think you need this any more now.

src/lib.rs
```diff
@@ -239,7 +308,60 @@ impl Client {
...
+            webdriver::capabilities::Capabilities::new(),
+            connector,
+        )
+            .await
+    }
+
+    /// Create a new `Client` associated with a chosen HttpsConnector and a new WebDriver session on the server
+    /// at the given URL.
+    ///
+    ///
```
jonhoo, line 343: Let's get rid of all these empty doc lines all over the place. They aren't rendered in the documentation anyway, and they make the comments harder to read :)

src/lib.rs
```diff
@@ -667,7 +790,7 @@ impl Client {
     }
 
     /// Switches to the frame specified at the index.
-    pub async fn enter_frame(mut self, index: Option<u16>) -> Result<Client, error::CmdError> {
+    pub async fn enter_frame(mut self, index: Option<u16>) -> Result<session::Client<C>, error::CmdError> {
```
jonhoo, line 793: Can't this just be `Client<C>`? Given how often we refer to `Client` in this file, I think we should just continue importing it the way we used to.

src/session.rs
jonhoo: Thinking some more about this, I think `Session` should only have a `with_capabilities_and_connector` constructor. The `Client` should just always be the one to create the connector and pass it in.

tests/common.rs
```diff
@@ -3,13 +3,20 @@
...
 
 use std::future::Future;
 use std::net::{IpAddr, Ipv4Addr, SocketAddr};
 use std::path::PathBuf;
 use warp::Filter;
 
+#[cfg(feature = "rustls-tls")]
+pub type Client = fantoccini::RustlsClient;
+#[cfg(all(feature = "openssl-tls", not(feature = "rustls-tls")))]
+pub type Client = fantoccini::OpenSslClient;
```
jonhoo, line 16: This won't work. We should run all the tests with both clients, not just with one or the other depending on features.

src/lib.rs
x0f5c3: That would work also, but I think that the type aliases help the user choose the TLS connector without specifying `hyper` and `hyper_rustls` or `hyper_tls` as their dependencies so even without the `NewConnector` trait I still think the type aliases are useful.

jonhoo: Ah, you mean if they store a `Client<...>` in a struct of some kind, and therefore name the type? That's a good point — let's keep them!

src/session.rs
x0f5c3: So we drop `with_capabilities` from `Session`, `new` and `with_capabilities` from generic `Client` and just do something like this?

```rust
#[cfg(feature = "rustls-tls")]
impl RustlsClient {
    pub async fn new_rustls(webdriver: &str) -> Result<Self, error::NewSessionError> {
        Client::new_with_connector(webdriver,hyper_rustls::HttpsConnector::new()).await
    }
   
    pub async fn rustls_with_capabilities(webdriver: &str, cap: webdriver::capabilities::Capabilities) -> Result<Self, error::NewSessionError> {
        Client::with_capabilities_and_connector(webdriver, cap, hyper_rustls::HttpsConnector::new()).await
    }
}
```
And another one for OpenSSL?
This way we can remove the `NewConnector` trait and just have seperate impls for Rustls and Openssl and the user can do just:
`Client::new_rustls("http://localhost:4444").await?`

jonhoo: Exactly! Although I would still keep `with_capabilities_and_connector` as a public function on `Client<C>` for anyone who wants to use their own connector type.

x0f5c3:
I still need to do tests for both features and improve docs to let the user know how to use the connectors but the `new_rustls` and the one for OpenSSL should be good now

src/session.rs
```diff
@@ -435,11 +391,73 @@ impl Session {
                 Ok(Client {
                     tx,
                     is_legacy: true,
+                    _marker: Default::default(),
                 })
             }
             Err(e) => Err(e),
         }
     }
+    fn shutdown(&mut self, ack: Option<Ack>) {
```
jonhoo, line 400: Could you please move this and `map_handshake_response` back up above to where they are on `master` just so that the diff is easier to read?

src/traits.rs
```diff
@@ -0,0 +1,22 @@
+
+/// Convenient methods for webdriver capabilities
+pub trait CapabilitiesExt {
```
jonhoo, line 3: I think maybe we should make this change in a separate PR, just so that they are easier to review. What do you think?

tests/common.rs
jonhoo: This is still relevant I believe -- we can't provide these aliases.

jonhoo:
This is starting to look really good! I've left a bunch of comments inline :)

If we want to really make this great, I think we should move to a [builder pattern](https://rust-lang.github.io/api-guidelines/type-safety.html#c-builder) for the client constructor now. Something like:
```rust
#[derive(Default, Clone)]
pub struct WebdriverHost<C> {
  capabilities: Option<...>,
  connector: C,
}

#[cfg(feature = "openssl-tls")]
impl WebDriverHost<hyper_tls::HttpsConnector> {
    pub fn openssl() -> Self { Self::new(..) }
}

#[cfg(feature = "rust-tls")]
impl WebDriverHost<hyper_rustls::HttpsConnector> {
    pub fn rustls() -> Self { Self::new(..) }
}

impl<C> WebDriverHost<C> {
    pub fn new(c: C) -> Self { .. }
    
    pub fn capabilities(&mut self, ...) -> &mut Self { .. }
    
    pub fn connect(&self) -> Result<Client<C>, ...> { .. }
}
```

I think that'll make the API much nicer!

src/traits.rs
x0f5c3: Ok, we can do that. That should make it easier for others to see the changes, I'll remove it from this.

x0f5c3:
> 
> 
> This is starting to look really good! I've left a bunch of comments inline :)
> 
> If we want to really make this great, I think we should move to a [builder pattern](https://rust-lang.github.io/api-guidelines/type-safety.html#c-builder) for the client constructor now. Something like:
> 
> ```rust
> #[derive(Default, Clone)]
> pub struct WebdriverHost<C> {
>   capabilities: Option<...>,
>   connector: C,
> }
> 
> #[cfg(feature = "openssl-tls")]
> impl WebDriverHost<hyper_tls::HttpsConnector> {
>     pub fn openssl() -> Self { Self::new(..) }
> }
> 
> #[cfg(feature = "rust-tls")]
> impl WebDriverHost<hyper_rustls::HttpsConnector> {
>     pub fn rustls() -> Self { Self::new(..) }
> }
> 
> impl<C> WebDriverHost<C> {
>     pub fn new(c: C) -> Self { .. }
>     
>     pub fn capabilities(&mut self, ...) -> &mut Self { .. }
>     
>     pub fn connect(&self) -> Result<Client<C>, ...> { .. }
> }
> ```
> 
> I think that'll make the API much nicer!

Great idea! I'll get on that, that should also simplify tests. I've been thinking and maybe the `tester!` macro could be modified to take an argument specifying the connector type.

x0f5c3:
I added the builder pattern, moved the code in session.rs,  and removed `CapabilitiesExt` , still need to figure out tests, I tried playing with the macro but the generic types make it a whole lot harder so I need some more time for that.

src/lib.rs
```diff
@@ -150,6 +170,76 @@ pub mod error;
...
+    /// # Arguments
+    /// * `connector` - TLS connector for [`Client`][hyper::client::Client]
+    pub fn new(connector: C) -> Self {
+        Self {
+            capabalities: None,
+            connector,
+        }
+    }
+    /// Set chosen webdriver capabilities
+    pub fn capabilities(mut self, cap: webdriver::capabilities::Capabilities) -> Self {
```
jonhoo, line 227: I think the recommendation is generally for builder methods to take `&mut self` and return `&mut Self`.

src/lib.rs
```diff
@@ -150,6 +170,76 @@ pub mod error;
...
+        self.capabalities = Some(cap);
+        self
+    }
+    /// Connect to the webdriver session
+    /// # Arguments
+    /// * `webdriver` - webdriver url
+    pub async fn connect(self, webdriver: &str) -> Result<Client<C>, error::NewSessionError> {
```
jonhoo, line 234: I think the recommendation is generally for the build method to take `&self` where possible. And I think that should be possible here? You may need to modify `Session::with_capabilities_and_connector` to take `&Capabilities`.

src/lib.rs
```diff
@@ -150,6 +170,76 @@ pub mod error;
...
+        return if let Some(cap) = self.capabalities {
```
jonhoo, line 235: You don't need the `return` here -- in Rust the last statement is automatically returned.

src/lib.rs
```diff
@@ -150,6 +170,76 @@ pub mod error;
...
+
+pub use session::Client;
+
+/// Builder pattern for Client
+#[derive(Default, Clone, Debug)]
+pub struct ClientBuilder<C>
+where
+    C: connect::Connect + Send + Sync + Clone + Unpin
+{
+    capabalities: Option<webdriver::capabilities::Capabilities>,
```
jonhoo, line 193: Typo: `capabilities`

tests/common.rs
```diff
@@ -3,20 +3,26 @@
...
+
 pub async fn select_client_type(s: &str) -> Result<Client, error::NewSessionError> {
     match s {
         "firefox" => {
             let mut caps = serde_json::map::Map::new();
             let opts = serde_json::json!({ "args": ["--headless"] });
             caps.insert("moz:firefoxOptions".to_string(), opts.clone());
-            Client::with_capabilities("http://localhost:4444", caps).await
+            #[cfg(feature = "rustls-tls")]
+            Client::rustls_with_capabilities("http://localhost:4444", caps).await
```
jonhoo, line 25: This probably has to be `Client::with_capabilities_and_connector`.

tests/common.rs
```diff
@@ -35,8 +41,8 @@ pub async fn select_client_type(s: &str) -> Result<Client, error::NewSessionErro
                     }
             });
             caps.insert("goog:chromeOptions".to_string(), opts.clone());
-
-            Client::with_capabilities("http://localhost:9515", caps).await
+            #[cfg(feature = "rustls-tls")]
+            Client::rustls_with_capabilities("http://localhost:9515", caps).await
```
jonhoo, line 45: And same here.

x0f5c3:
I fixed the tests to do both connectors but maybe it's an overcomplicated solution. I changed the builder pattern to take `&mut self` and fixed the docs. If this is ok, I can get to fixing conflicts with the new version based on new `hyper` and `tokio 1.0`

src/lib.rs
```diff
@@ -219,23 +215,23 @@ where
...
         self
     }
     /// Connect to the webdriver session
     /// # Arguments
     /// * `webdriver` - webdriver url
-    pub async fn connect(self, webdriver: &str) -> Result<Client<C>, error::NewSessionError> {
-        return if let Some(cap) = self.capabalities {
-            Client::with_capabilities_and_connector(webdriver, cap, self.connector).await
+    pub async fn connect(&self, webdriver: &str) -> Result<Client<C>, error::NewSessionError> {
+        if let Some(cap) = self.capabilities.clone() {
```
jonhoo, line 231: You shouldn't need the `.clone()` here I don't think — you just need to write either:
```rust
if let Some(ref cap) = self.capabilities {
```
or
```rust
if let Some(cap) = &self.capabilities {
```

src/lib.rs
```diff
@@ -321,7 +317,7 @@ where
     ) -> Result<Self, error::NewSessionError> {
         Self::with_capabilities_and_connector(
             webdriver,
-            webdriver::capabilities::Capabilities::new(),
+            &mut webdriver::capabilities::Capabilities::new(),
```
jonhoo, line 320: Shouldn't this just be `&`?

src/lib.rs
```diff
@@ -348,10 +344,10 @@ where
     /// `Client` is dropped.
     pub async fn with_capabilities_and_connector(
         webdriver: &str,
-        cap: webdriver::capabilities::Capabilities,
+        cap: &webdriver::capabilities::Capabilities,
         connector: C,
     ) -> Result<Self, error::NewSessionError> {
-        Session::with_capabilities_and_connector(webdriver, cap, connector).await
+        Session::with_capabilities_and_connector(webdriver, cap.to_owned(), connector).await
```
jonhoo, line 350: Ah, I'm proposing you make `Session::with_...` _also_ take `&Capabilities`.

tests/common.rs
```diff
@@ -9,20 +9,49 @@ use std::future::Future;
...
+                        "/usr/bin/chromium-browser"
+                    } else if std::path::Path::new("/Applications/Google Chrome.app/Contents/MacOS/Google Chrome").exists() {
+                        // macOS
+                        "/Applications/Google Chrome.app/Contents/MacOS/Google Chrome"
+                    } else {
+                        // elsewhere, it's just called chromium
+                        "/usr/bin/chromium"
+                    }
+            });
+            caps.insert("goog:chromeOptions".to_string(), opts.clone());
```
jonhoo, line 40: Let's try to put the (shared) code for constructing the capabilities into its own function which the two TLS configurations can then call out to.

tests/common.rs
```diff
@@ -41,13 +70,13 @@ pub async fn select_client_type(s: &str) -> Result<Client, error::NewSessionErro
...
             caps.insert("goog:chromeOptions".to_string(), opts.clone());
-            #[cfg(feature = "rustls-tls")]
-            Client::rustls_with_capabilities("http://localhost:9515", caps).await
+            ClientBuilder::openssl().capabilities(caps).connect("http://localhost:9515").await
         }
         browser => unimplemented!("unsupported browser backend {}", browser),
     }
 }
 
+
```
jonhoo, line 79: This seems like a stray line.

tests/common.rs
```diff
@@ -70,13 +99,15 @@ pub fn handle_test_error(
...
+#[cfg(feature = "rustls-tls")]
 #[macro_export]
-macro_rules! tester {
+macro_rules! rustls_tester {
     ($f:ident, $endpoint:expr) => {{
         use std::sync::{Arc, Mutex};
         use std::thread;
 
-        let c = common::select_client_type($endpoint);
+        let c = common::rustls_type($endpoint);
```
jonhoo, line 110: Let's avoid the duplication here, and have the macro take `$connector:expr` instead of `$endpoint:expr`. And then make this line be
```rust
let c = $connector;
```
Then, rename this macro to `tester_inner`, and add another macro like this:
```rust
macro_rules! tester {
    ($f:ident, $endpoint:expr) => {{
        #[cfg(feature = "rustls-tls")]
        tester_inner!($f, rustls_type($endpoint));
        #[cfg(feature = "openssl-tls")]
        tester_inner!($f, opentls_type($endpoint));
}
```
That way, all the tests can stay the same, and they'll test both TLS implementations.

tests/common.rs
```diff
@@ -107,12 +138,62 @@ macro_rules! tester {
...
+
+#[cfg(feature = "openssl-tls")]
 #[macro_export]
-macro_rules! local_tester {
+macro_rules! openssl_local {
     ($f:ident, $endpoint:expr) => {{
         let port: u16 = common::setup_server();
-        let f = move |c: Client| async move { $f(c, port).await };
-        tester!(f, $endpoint)
+        let f = move |c: Client<hyper_tls::HttpsConnector<hyper::client::HttpConnector>>| async move { $f(c, port).await };
```
jonhoo, line 195: I _think_ this can just be
```rust
let f = move |c| async move { $f(c, port).await };
```

src/lib.rs
x0f5c3: `Session` constructor modifies the capabilities so somewhere it would need to be `&mut Capabilities` from what I can see

jonhoo: Ah, I would just have the `Session` constructor do:
```rust
let mut capabilities = cap.to_owned();
```
internally

tests/common.rs
x0f5c3: Unfortunately the only way I found to do this was to create the closure twice for each connector because it needs a concrete type.

tests/common.rs
```diff
@@ -38,45 +38,34 @@ pub async fn rustls_type(s: &str) -> Result<Client<hyper_rustls::HttpsConnector<
...
+            caps
         }
         browser => unimplemented!("unsupported browser backend {}", browser),
     }
 }
 
+#[cfg(feature = "rustls-tls")]
+pub async fn rustls_type(s: &str) -> Result<Client<hyper_rustls::HttpsConnector<hyper::client::HttpConnector>>, error::NewSessionError> {
+    let caps = make_capabilities(s);
+    let url = match s {
```
jonhoo, line 50: This `match` should also be factored out so we don't have to repeat it for each backend type.

tests/common.rs
```diff
@@ -35,13 +38,34 @@ pub async fn select_client_type(s: &str) -> Result<Client, error::NewSessionErro
...
-
-            Client::with_capabilities("http://localhost:9515", caps).await
+            caps
         }
         browser => unimplemented!("unsupported browser backend {}", browser),
     }
 }
 
+#[cfg(feature = "rustls-tls")]
+pub async fn rustls_type(s: &str) -> Result<Client<hyper_rustls::HttpsConnector<hyper::client::HttpConnector>>, error::NewSessionError> {
```
jonhoo, line 48: Should these be called `rustls_client`/openssl_client` instead? The `_type` suffix is weird to me here.

tests/local.rs
```diff
@@ -4,23 +4,33 @@ extern crate serial_test_derive;
 extern crate fantoccini;
 extern crate futures_util;
 
-use fantoccini::{error, Client, Locator};
+use hyper::client::connect::Connect;
+
+use fantoccini::{error, Locator};
+
+use fantoccini::Client;
```
jonhoo, line 11: Let's put this back with the other imports further up the way it was before so we simplify the diff

tests/local.rs
```diff
@@ -332,162 +381,164 @@ async fn resolve_execute_async_value(mut c: Client, port: u16) -> Result<(), err
 
 mod firefox {
     use super::*;
+    use common::*;
     #[test]
     #[serial]
     fn navigate_to_other_page() {
-        local_tester!(goto, "firefox")
+        tester_local!(goto, "firefox");
```
jonhoo, line 388: These diffs seem unnecessary. Can we just have the macro be called `tester_local` the way it was?

tests/local.rs
```diff
@@ -332,162 +381,164 @@ async fn resolve_execute_async_value(mut c: Client, port: u16) -> Result<(), err
 
 mod firefox {
     use super::*;
+    use common::*;
```
jonhoo, line 384: Hmm, why do we need this `use` now? If it's because of something we use in the macro, we can just stick the `use` in the macro directly instead.

jonhoo:
I think we're getting really close with this now — great work! Sorry to keep pestering you with comments, I just want to make sure we land this in the best way we can :) Thank you for sticking with it.

x0f5c3:
No problem, I understand. I tried to reduce function calls and cloning so it should be cleaner now.

x0f5c3:
I moved everything to tokio 1.0 and new hyper versions but now it fails on `it_can_be_raw` tests and to be honest I have no idea why.

jonhoo:
You probably need to merge with the latest `master`. That might resolve the issues you're having.

x0f5c3:
It fails only with `rustls` enabled and with `Failed(hyper::Error(IncompleteMessage)` error, I tried to find out why it failed but I can't. It works with only openssl enabled. I think all the changes from `master` are merged except where they're required to use rustls

jonhoo:
Hmm, GitHub still complains that you have conflicting files. Have you done a `git merge master` with the latest `master`?

x0f5c3:
I merged now and tests still fail, I'm going to try some things, see if I can find why it fails

x0f5c3:
It works with `http2` feature enabled on `hyper`. Also I think we could drop the generics on `Client`, `Element`, `Form`, now that we have the builder pattern, it should simplify the whole thing for the end user. I'm using this in a project of mine and the generics really polluted my code. I also bumped `url` and `mime` versions but I can take it back if needed. I don't know if the `http2` would add a lot of bloat to the library, it only pulls `h2` and it doesn't have a lot of dependencies that aren't already in other libraries that we use.

jonhoo:
It's _really_ weird for `http2` to be required for this to function — it suggests that the driver _only_ supports `http2`. Does it not work with the `http1` feature instead?

I'm not sure I follow how we'd remove the generics from `Client` and the other types?

Also, there are still a lot of stray blank lines in the diff — could you remove those again?

x0f5c3:
No it doesn't work with only `http1`. I couldn't really find a reason why `http2` solves it, the closest I came was some mentions that it's about the `http1` timeout but I don't know really why it doesn't work with rustls and works with openssl, it started happening after upgrading to tokio 1.0 and new hyper. Does the `http2` feature pull too much dependencies?

I'll remove the blank lines, sorry for that. 

Basically since the Client and other types don't hold a `Session` or `hyper::Client` inside we can create both types of `Session` just by the builder pattern or manually by `new_with_connector`  and that allows us to remove the generic types from the user-facing types which will make using the library easier since there won't be a need to specify `Element<HttpsConnector<HttpConnector>>` and the same for `Form`, `Client` etc. 
I found that using the library is a lot harder with the generics since if I want to make a function that takes a `Client` or `Element` or something like that I need to specify the generics and also IDEs and LSPs can bug out with type elision. 
So the user just needs to do `ClientBuilder::rustls().connect("http://localhost:4444").await?;`

src/lib.rs
```diff
@@ -212,21 +283,38 @@ pub struct Form {
...
+        .await
     }
 
-    /// Create a new `Client` associated with a new WebDriver session on the server at the given
-    /// URL.
+    /// Create a new `Client` associated with a chosen HttpsConnector and a new WebDriver session on the server
+    /// at the given URL.
+    ///
+    ///
+    ///
```
jonhoo, line 311: Extra newlines here.

jonhoo:
No, `http2` is fine, I'm just really curious why it makes a difference :sweat_smile: 

Ah, no, I'm referring to a couple of stray newlines that _do_ show up the code itself. I'll go through and leave line comments by them.

Oh, you're right! I had it in my head that the `Client` contains a `Session`, but it doesn't, it goes through a channel. Good catch! Yes, absolutely, removing the generics from all the `Client`/`Element`/`Form` types would be a huge win!

src/lib.rs
```diff
@@ -13,6 +13,11 @@
 //! For low-level access to the page, `Client::source` can be used to fetch the full page HTML
 //! source code, and `Client::raw_client_for` to build a raw HTTP request for a particular URL.
 //!
+//! # Feature flags
+//! - `rustls-tls`: Use rustls for https, enabled by default
+//! - `openssl-tls`: Use openssl for https
+//!
+//!
```
jonhoo, line 20: Extra newline here

src/lib.rs
```diff
@@ -193,8 +266,6 @@ impl<'a> From<Locator<'a>> for webdriver::command::LocatorParameters {
     }
 }
 
-pub use crate::session::Client;
```
jonhoo, line 196: Why did this have to move?

src/lib.rs
```diff
@@ -212,21 +283,38 @@ pub struct Form {
 }
 
 impl Client {
-    /// Create a new `Client` associated with a new WebDriver session on the server at the given
-    /// URL.
+    /// Create a new [`Client`][crate::Client] associated with a chosen HttpsConnector and a new WebDriver session on the server
+    /// at the given URL.
+    ///
```
jonhoo, line 288: Extra newline here

src/lib.rs
jonhoo, line 311: Extra newlines here

src/lib.rs
```diff
@@ -212,21 +283,38 @@ pub struct Form {
...
+    ///
+    ///
     ///
+    /// Calls `with_capabilities_and_connector` with an empty capabilities list.
     /// The given capabilities will be requested in `alwaysMatch` or `desiredCapabilities`
     /// depending on the protocol version supported by the server.
     ///
+    ///
```
jonhoo, line 317: Extra newline here.

tests/common.rs
```diff
@@ -105,15 +152,42 @@ macro_rules! tester {
...
         let f = move |c: Client| async move { $f(c, port).await };
-        tester!(f, $endpoint)
+        #[cfg(feature = "rustls-tls")]
+        tester_inner!(f, common::make_client(url, caps.clone(), "rustls"));
+        #[cfg(feature = "openssl-tls")]
+        tester_inner!(f, common::make_client(url, caps, "openssl"))
     }};
 }
 
+// #[cfg(feature = "openssl-tls")]
```
jonhoo, line 170: Why did this go away?

tests/local.rs
```diff
@@ -7,7 +7,6 @@ extern crate futures_util;
 use fantoccini::{error, Client, Locator};
 
 mod common;
-
```
jonhoo, line 10: Newline accidentally deleted

tests/local.rs
```diff
@@ -335,159 +334,159 @@ mod firefox {
...
     #[serial]
     fn select_by_label_test() {
-        local_tester!(select_by_label, "chrome")
+        local_tester!(select_by_label, "chrome");
     }
 
     #[test]
     fn select_by_index_label() {
-        local_tester!(select_by_label, "chrome")
+        local_tester!(select_by_index, "chrome");
```
jonhoo, line 490: Good catch!

tests/remote.rs
```diff
@@ -258,157 +258,155 @@ async fn wait_for_navigation_test(mut c: Client) -> Result<(), error::CmdError>
 
 mod chrome {
     use super::*;
-
```
jonhoo, line 261: Accidentally removed newline

tests/remote.rs
```diff
@@ -258,157 +258,155 @@ async fn wait_for_navigation_test(mut c: Client) -> Result<(), error::CmdError>
...
     #[test]
     fn it_waits_for_navigation() {
-        tester!(wait_for_navigation_test, "chrome")
+        tester!(wait_for_navigation_test, "chrome");
     }
 }
 
 mod firefox {
     use super::*;
-
```
jonhoo, line 336: Accidentally removed newline

src/lib.rs
x0f5c3: Sorry, I removed some time while doing something it that part of the file and then when I inserted it again I just did it at the top of the file like I do all imports, but I moved it now to where it's been

tests/common.rs
x0f5c3: I don't know why I commented it out since we need just one local_tester, sorry about that. I fixed it now

src/lib.rs
```diff
@@ -150,6 +166,61 @@ pub mod error;
...
+    pub fn openssl() -> Self {
+        Self::new(hyper_tls::HttpsConnector::new())
+    }
+}
+impl<C> ClientBuilder<C>
+where
+    C: connect::Connect + Send + Sync + Clone + Unpin + 'static,
+{
+    /// Create a new builder with TLS connector
+    /// # Arguments
```
jonhoo, line 199: In Rust, arguments aren't usually documented as lists (unlike in Java for example). Instead, they are more commonly incorporated into explanatory text. I'd suggest:
```rust
/// Create a builder using `connector` as the TLS connector.
```

src/lib.rs
```diff
@@ -150,6 +166,61 @@ pub mod error;
...
+    /// * `connector` - TLS connector for [`Client`][hyper::client::Client]
+    pub fn new(connector: C) -> Self {
+        Self {
+            capabilities: None,
+            connector,
+        }
+    }
+    /// Set chosen webdriver capabilities
```
jonhoo, line 207: Not sure what you mean by "chosen" here? Can this just be
```rust
/// Pass the given WebDriver capabilities to the browser.
```

jonhoo: Also, in Rust, the convention is generally to leave an empty line between functions. So a blank line before this line.

src/lib.rs
```diff
@@ -150,6 +166,61 @@ pub mod error;
...
+    pub fn capabilities(&mut self, cap: webdriver::capabilities::Capabilities) -> &mut Self {
+        self.capabilities = Some(cap);
+        self
+    }
+    /// Connect to the webdriver session
+    /// # Arguments
```
jonhoo, line 213: Same here re documenting arguments in Rust.

src/lib.rs
```diff
@@ -212,18 +283,29 @@ pub struct Form {
 }
 
 impl Client {
-    /// Create a new `Client` associated with a new WebDriver session on the server at the given
-    /// URL.
+    /// Create a new [`Client`][crate::Client] associated with a chosen HttpsConnector and a new WebDriver session on the server
```
jonhoo, line 286: You should not need to give `[crate::Client]` here — rustdoc will figure that automatically on recent compiler versions.

Also, in Rust, the first line of documentation for a function is usually a very short summary of what the function does, with subsequent paragraphs going into more detail. I'd suggest:
```rust
/// Connect to the WebDriver host running the given address.
///
/// The provided `connector` is used to establish the connection to the WebDriver host, and should generally be one that supports HTTPS, as that is commonly required by WebDriver implementations.
```

(though line-wrapped appropriately)

src/lib.rs
```diff
@@ -212,18 +283,29 @@ pub struct Form {
...
+            &webdriver::capabilities::Capabilities::new(),
+            connector,
+        )
+        .await
     }
 
-    /// Create a new `Client` associated with a new WebDriver session on the server at the given
-    /// URL.
+    /// Create a new `Client` associated with a chosen HttpsConnector and a new WebDriver session on the server
+    /// at the given URL.
```
jonhoo, line 306: Same note about docs here as above.

src/lib.rs
```diff
@@ -212,18 +283,29 @@ pub struct Form {
...
     ///
+    /// Calls `with_capabilities_and_connector` with an empty capabilities list.
```
jonhoo, line 308: This should be removed as this _is_ `with_capabilities_and_connector`.

src/lib.rs
```diff
@@ -1149,6 +1235,7 @@ impl Element {
         let Self {
             mut client,
             element,
+            ..
```
jonhoo, line 1238: I don't think this change is needed any more? I'd rather keep it without `..` so that we remember to update it should we add fields in the future.

src/session.rs
```diff
@@ -328,20 +337,18 @@ impl Session {
             }
         }
     }
-
-    pub(crate) async fn with_capabilities(
+    pub(crate) async fn with_capabilities_and_connector(
```
jonhoo, line 340: As noted earlier, please keep the blank line between functions.

src/session.rs
```diff
@@ -328,20 +337,18 @@ impl Session {
...
-        mut cap: webdriver::capabilities::Capabilities,
+        cap: &webdriver::capabilities::Capabilities,
+        connector: C,
     ) -> Result<Client, error::NewSessionError> {
         // Where is the WebDriver server?
         let wdb = webdriver.parse::<url::Url>();
-
         let wdb = wdb.map_err(error::NewSessionError::BadWebdriverUrl)?;
-
         // We want a tls-enabled client
```
jonhoo, line 348: This comment is no longer accurate and should be removed.

src/session.rs
```diff
@@ -358,7 +365,7 @@ impl Session {
         });
 
         // now that the session is running, let's do the handshake
-        let mut client = Client {
+        let mut client: Client = Client {
```
jonhoo, line 368: I don't think this change is necessary?

src/session.rs
```diff
@@ -328,20 +337,18 @@ impl Session {
...
         let wdb = webdriver.parse::<url::Url>();
-
         let wdb = wdb.map_err(error::NewSessionError::BadWebdriverUrl)?;
-
         // We want a tls-enabled client
-        let client = hyper::client::Client::builder()
-            .build::<_, hyper::Body>(hyper_tls::HttpsConnector::new());
+        let client = hyper::Client::builder().build::<_, hyper::Body>(connector);
 
+        let mut caps = cap.to_owned();
```
jonhoo, line 351: We could just make this
```rust
let mut cap = cap.to_owned();
```
and that way much of the code below can stay the same.

src/session.rs
```diff
@@ -422,7 +429,7 @@ impl Session {
                 // WebDriver protocol:
                 // https://github.com/SeleniumHQ/selenium/wiki/JsonWireProtocol
                 let session_config = webdriver::capabilities::LegacyNewSessionParameters {
-                    desired: cap,
+                    desired: caps.to_owned(),
```
jonhoo, line 432: I don't think this needs to call `to_owned` any more?

tests/common.rs
```diff
@@ -3,10 +3,13 @@
 extern crate fantoccini;
 extern crate futures_util;
 
-use fantoccini::{error, Client};
+use fantoccini::error;
```
jonhoo, line 7: Let's just remove this line here and merge the `use` blocks for `fantoccini` into:
```rust
use fantoccini::{error, Client, ClientBuilder};
```

tests/common.rs
```diff
@@ -40,9 +43,40 @@ pub async fn select_client_type(s: &str) -> Result<Client, error::NewSessionErro
...
+                .await
         }
+        #[cfg(feature = "openssl-tls")]
+        "openssl" => {
+            ClientBuilder::openssl()
+                .capabilities(caps)
+                .connect(url)
+                .await
+        }
+        other => unimplemented!("Unsupported connector type {}", other),
```
jonhoo, line 72: To make debugging a little easier, I would include:
```rust
#[cfg(not(feature = "rustls-tls"))]
"rusttls" => panic!("Asked to run rusttls test, but rustls-tls feature is not enabled"),
#[cfg(not(feature = "openssl-tls"))]
"openssl" => panic!("Asked to run openssl test, but openssl-tls feature is not enabled"),
```

jonhoo:
This is looking awesome now! I left what I think will probably be the last set of change comments :tada:

x0f5c3:
It should be good now, but if I forgot something, let me know.

src/lib.rs
```diff
@@ -195,23 +195,19 @@ impl<C> ClientBuilder<C>
...
-    /// Set chosen webdriver capabilities
+    /// Pass the given WebDriver capabilities to the browser.
     pub fn capabilities(&mut self, cap: webdriver::capabilities::Capabilities) -> &mut Self {
         self.capabilities = Some(cap);
         self
     }
-    /// Connect to the webdriver session
-    /// # Arguments
-    /// * `webdriver` - webdriver url
+    /// Connect to the webdriver session at the `url`
```
jonhoo, line 210: There isn't an argument called `url` here?

Cargo.toml
```diff
@@ -17,19 +17,25 @@ categories = ["api-bindings", "development-tools::testing", "web-programming::ht
...
 
+[features]
+default = ["rustls-tls"]
+rustls-tls = ["hyper-rustls"]
+openssl-tls = ["hyper-tls"]
+
 [dependencies]
 webdriver = { version = "0.43.0", default-features = false }
-url = "2.0.0"
+url = "2.2.0"
```
jonhoo, line 27: Is this version bump required?

Cargo.toml
```diff
@@ -17,19 +17,25 @@ categories = ["api-bindings", "development-tools::testing", "web-programming::ht
...
 serde = { version = "1.0", features = [ "derive" ] }
 serde_json = "1.0"
-futures-core = "0.3.0"
-futures-util = "0.3.0"
+futures-core = "0.3.12"
+futures-util = "0.3.12"
```
jonhoo, line 31: Are these version bumps required?

Cargo.toml
```diff
@@ -17,19 +17,25 @@ categories = ["api-bindings", "development-tools::testing", "web-programming::ht
...
 tokio = { version = "1", features = [ "sync", "rt" ] }
-hyper = { version = "0.14", features = [ "stream", "client", "http1" ] }
-hyper-tls = "0.5.0"
+hyper = { version = "0.14", features = [ "stream", "client", "http1", "http2" ] }
 cookie = { version = "0.14", features = ["percent-encode"] }
 base64 = "0.13"
-mime = "0.3.9"
+hyper-rustls = { version = "0.22.1", optional = true }
+hyper-tls = { version = "0.5.0", optional = true }
+mime = "0.3.16"
```
jonhoo, line 38: Is this version bump required?

x0f5c3:
The bumps were not required, I just saw that they can be done and thought that maybe it would be better to keep dependencies up to date but I downgraded them back to the ones that are on master.

Issue #114, opened 2020-12-21
Update to tokio 1.0


MonsieurMan:
Hi there !

I'll be using your library for a little raspberry project, checking tokyo I see there is a new version.

I'm in need of a little more training in Rust, would you be interested in me landing a PR for a migration to tokyo 0.3, or are there any current work on the matter ?

Thanks in advance,

jonhoo:
Hi! Absolutely, I'd love one. I think we're probably going to be blocked on `hyper` not yet supporting `tokio 0.3`, but getting the ball rolling is a good idea :)

MonsieurMan:
`hyper` author is going to wait a few weeks for `tokio 1.0` to ship, in order not to burden his users with potentially two breaking releases.

jonhoo:
Now that `tokio 0.1` and `hyper 0.14` are out, I think we're only blocked on `hyper-tls` to get this out the door (https://github.com/hyperium/hyper-tls/issues/76)!

jonhoo:
See also https://bugzilla.mozilla.org/show_bug.cgi?id=1684226

bnoctis:
And [`tokio` 1.0 was released](https://tokio.rs/blog/2020-12-tokio-1-0). What a pace for them.

peppizza:
hyper-tls v0.5 has been released with tokio 1.0, so if that was the last thing blocking then we should be free to update this lib

jonhoo:
We unfortunately also currently have `warp` as a dev-dependency, which is still on `tokio 0.2` (issue here: https://github.com/seanmonstar/warp/issues/725). If someone wants to submit a PR that moves the tests to use `hyper` for the server instead (shouldn't be _too_ painful), then that would unblock the move!

jonhoo:
https://github.com/jonhoo/fantoccini/pull/118

PR #117, opened 2021-01-06
Replace the test warp server with hyper (remove warp as a dependency)

soerenmeier:
As suggested by @jonhoo in https://github.com/jonhoo/fantoccini/issues/114#issuecomment-755096495 to upgrade faster to tokio 1.0 we could remove warp as a dependency.
The warp server gets replaced by a very simple hyper service. As I don't see a reason why test files could be of another type than html, I've hard coded that in. But this could be fixed quite easily if at any point in time that assumption changes.
Or I can edit the pull request to be more "generic" over the file type, but that would probably require another dependency.

I would be happy to hear some feedback, this is my first pull request ever.

### DIFF
--- a/tests/common.rs
+++ b/tests/common.rs
@@ -7,11 +7,13 @@ use fantoccini::{error, Client};
 
 use std::future::Future;
 use std::net::{IpAddr, Ipv4Addr, SocketAddr};
-use std::path::PathBuf;
+use std::path::Path;
 use std::convert::Infallible;
 use hyper::service::{make_service_fn, service_fn};
 use hyper::{Server, Request, Response, Body, StatusCode};
 use tokio::fs::read_to_string;
+
+const ASSETS_DIR: &str = "tests/test_html";
 
 pub async fn select_client_type(s: &str) -> Result<Client, error::NewSessionError> {
     match s {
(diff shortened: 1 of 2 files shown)


jonhoo:
This looks great, thank you! There's one change I'd like to see (comment below), but apart from that this looks :+1:

tests/common.rs
```diff
@@ -124,27 +127,67 @@ pub fn setup_server() -> u16 {
...
 
     rx.recv().expect("To get the bound port.")
 }
 
 /// Configures and starts the server
-fn start_server() -> (SocketAddr, impl Future<Output = ()> + 'static) {
+fn start_server() -> (SocketAddr, impl Future<Output = hyper::Result<()>> + 'static) {
     let socket_addr = SocketAddr::new(IpAddr::V4(Ipv4Addr::LOCALHOST), 0);
+
     const ASSETS_DIR: &str = "tests/test_html";
```
jonhoo, line 141: Let's hoist this to the top of the file, and then use it directly from inside `handle_file_request` with:
```rust
let asset_file = Path::new(ASSETS_DIR).join(uri_path);
```

soerenmeier: Sounds good, that makes the service creation easier to.

jonhoo (approved):

jonhoo:
Excellent! Will do the bump to tokio 1.0 over the weekend :tada:

jonhoo:
https://github.com/jonhoo/fantoccini/pull/118

PR #126, opened 2021-02-25
Add error type for pageload timeout

### LINKED ISSUE #120: panicked at 'internal error: entered unreachable code: received unknown error (timeout)
When crawling a website, I get this when it happens upon a certain page:
```
thread 'tokio-runtime-worker' panicked at 
'internal error: entered unreachable code: received unknown error (timeout) for INTERNAL_SERVER_ERROR status code',
/home/spooder/.cargo/registry/src/github.com-1ecc6299db9ec823/fantoccini-0.15.0/src/session.rs:806:34
```
It also seems geckodriver dies at this point, as I'll get the following on the next pages.
```
webdriver connection lost: WebDriver session was closed while waiting
webdriver connection lost: WebDriver session has been closed
webdriver connection lost: WebDriver session has been closed
// etc
```
Is there a way to circumvent this error?  Anything I could do about it?

leaty:
Fixes #120.

As I wrote in the issue, the error occurred when `.goto()` was used on e.g. any URL redirecting to a download, which opens a download prompt (invisible when running headless), this seems to make geckodriver return with a timeout error because the "page" never loaded, i.e. the download. 

The error appeared after 5 minutes, which is exactly the default `pageLoad` timeout. After setting it to 5 seconds, it instead occurred *exactly* after 5 seconds. Based on this I believe this is the correct `ErrorStatus`, or at the very least this is better than dropping the connection without any ability to actually shutdown the still-alive session.

In any case, it works for me! One funny thing is that the download prompt remains forever instead of closing, but it keeps chugging along to the next page(s) and running all the code as usual. It could *possibly* be closed by using ActionChains and e.g. sending keydown Escape.

### DIFF
--- a/src/session.rs
+++ b/src/session.rs
@@ -807,6 +807,7 @@ where
                             "unknown error" => ErrorStatus::UnknownError,
                             "script timeout" => ErrorStatus::ScriptTimeout,
                             "unsupported operation" => ErrorStatus::UnsupportedOperation,
+                            "timeout" => ErrorStatus::Timeout,
                             _ => unreachable!(
                                 "received unknown error ({}) for INTERNAL_SERVER_ERROR status code",
                                 error


jonhoo:
Thanks for digging all the way through that! This is definitely a bug in the WebDriver implementation, because the [spec clearly states](https://www.w3.org/TR/webdriver1/#handling-errors) that a timeout error should be accompanied by a 408 Request Timeout code, which we already handle over here:
https://github.com/jonhoo/fantoccini/blob/8e134afa2cde1d114326f06d76536c18eb763049/src/session.rs#L816

That said, I'm happy to merge this as a workaround if it fixes your use-case! I'd also suggest reporting this as a bug upstream though :)

As for your problem with the prompts, you may find the [User Prompt actions](https://www.w3.org/TR/webdriver1/#user-prompts) handy! We don't currently expose those, but they should be easy enough to add!

jonhoo (approved):

jonhoo:
Released in `0.17.3` :tada:

leaty:
> [spec clearly states](https://www.w3.org/TR/webdriver1/#handling-errors) that a timeout error should be accompanied by a 408 Request Timeout code, which we already handle over here:

Ah! Yes, that makes more sense now, geckodriver handles this specific case out-of-spec.

> That said, I'm happy to merge this as a workaround if it fixes your use-case!

It took a lot of research and testing because I wanted to be as sure as possible, it really didn't make a lot of sense to me, yet the fix- is just one line, but with it the crawler works phenomenally now!

> Released in `0.17.3` :tada:

Awesome! :1st_place_medal:

PR #130, opened 2021-05-23
Add support for WebDriver cookie operations

### LINKED ISSUE #59: Request: Add an API for interacting w/ Cookies
_Full disclaimer: I'm not a web expert and I'm new to using WebDriver so this might be possible using functionality that lives outside of Fantoccini, so feel free to close if that's the case_

Currently, the Client does not support an easy way to interact with the current Session's cookie store unless you make a raw request and interact with the request builder/response which will only give you the cookies for that particular domain.

It would be nice to be able to interact with a Session's cookie store from the high level API to enable things like serializing/deserializing persistent cookies in applications where you may want to remember things like user log-in state. Examples of where this could be useful might include improving CI/CD test times by not requiring repeated log-ins. It could also be useful in scraper-backed user-facing applications that use technologies like OAuth instead of Basic Auth, so that the user doesn't have to repeatedly authorize the login every time they launch the application (even if it's automated).

aksiksi:
## Overview

This PR adds support for the cookie operations defined in [16. Cookies](https://www.w3.org/TR/webdriver1/#cookies) in the WebDriver spec. In particular, I've implemented the following operations:

* Get All Cookies
* Get Named Cookie
* Delete Cookie
* Delete All Cookies

The last operation, "Add Cookie", requires support from the `webdriver` crate.  The primary issue is that the `AddCookieParameters` struct does not implement `Serialize`, which means that we can't build the POST request on our side.  I'll be trying to get in touch with the geckodriver folks to see if this can be changed on their end.

## Changes

* Added a new module, `cookies.rs`.  This module contains the cookie-related `Client` methods as well as a few functions for converting between raw JSON (from WebDriver) and `cookie::Cookie` for the APIs.
* Added a simple integration testcase that triggers the `get_all_cookies` and `delete_all_cookies` operations.
* Removed the `binary` flag/arg when creating the WebDriver test client.  Both `chromedriver` and `geckodriver` are able to detect the path to the browser binary automatically on Windows and Linux.

Closes #59.

### DIFF
--- a/src/cookies.rs
+++ b/src/cookies.rs
@@ -1,0 +1,150 @@
+use serde_json::Value as Json;
+use webdriver::command::WebDriverCommand;
+
+use crate::client::Client;
+use crate::error;
+
+/// Key names for cookie fields used by WebDriver JSON.
+const COOKIE_NAME: &str = "name";
+const COOKIE_VALUE: &str = "value";
+const COOKIE_PATH: &str = "path";
+const COOKIE_DOMAIN: &str = "domain";
+const COOKIE_SECURE: &str = "secure";
+const COOKIE_HTTP_ONLY: &str = "httpOnly";
+const COOKIE_EXPIRY: &str = "expiry";
+
+/// Build a `cookie::Cookie` from raw JSON.
+fn json_to_cookie(raw_cookie: &serde_json::Map<String, Json>) -> cookie::Cookie<'static> {
+    // Required keys
+    let name = raw_cookie.get(COOKIE_NAME).and_then(|v| v.as_str()).unwrap().to_string();
+    let value = raw_cookie.get(COOKIE_VALUE).and_then(|v| v.as_str()).unwrap().to_string();
+
+    let mut cookie = cookie::Cookie::new(name, value);
+
+    // Optional keys
+    let path = raw_cookie.get(COOKIE_PATH).and_then(|v| v.as_str()).map(String::from);
+    let domain = raw_cookie.get(COOKIE_DOMAIN).and_then(|v| v.as_str()).map(String::from);
+    let secure = raw_cookie.get(COOKIE_SECURE).and_then(|v| v.as_bool());
+    let http_only = raw_cookie.get(COOKIE_HTTP_ONLY).and_then(|v| v.as_bool());
+    let expiry = raw_cookie.get(COOKIE_EXPIRY).and_then(|v| v.as_u64());
+
+    if let Some(path) = path {
+        cookie.set_path(path);
+    }
+
+    if let Some(domain) = domain {
+        cookie.set_domain(domain);
+    }
+
+    if let Some(secure) = secure {
+        cookie.set_secure(secure);
+    }
+
+    if let Some(http_only) = http_only {
+        cookie.set_http_only(http_only);
+    }
+
+    if let Some(_expiry) = expiry {
+        todo!()
+    }
+
+    cookie
+}
+
+/// Serialize a `cookie::Cookie` to JSON.
+#[allow(unused)]
+fn cookie_to_json(cookie: &cookie::Cookie<'_>) -> Json {
+    let mut json = serde_json::json!(
+        { COOKIE_NAME: cookie.name(), COOKIE_VALUE: cookie.value() }
+    );
+
+    if let Some(path) = cookie.path() {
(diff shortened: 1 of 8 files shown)


src/cookies.rs
```diff
@@ -0,0 +1,123 @@
...
+use webdriver::command::WebDriverCommand;
+
+use crate::client::Client;
+use crate::error;
+
+/// Type alias for a `cookie::Cookie`
+pub type Cookie<'a> = cookie::Cookie<'a>;
+
+/// JSON representation of a cookie as [defined by WebDriver](https://www.w3.org/TR/webdriver1/#cookies).
+#[derive(Debug, Deserialize, Serialize)]
```
aksiksi, line 13: It might make sense to expose this type directly instead of converting to/from a `cookie::Cookie`.  The problem with using `cookie::Cookie` is that some fields will not be forwarded to WebDriver, even when set by the user (e.g., `same_site`).

src/cookies.rs
```diff
@@ -0,0 +1,123 @@
...
+struct JsonCookie {
```
aksiksi, line 14: It might make sense to expose this type directly instead of converting to/from a cookie::Cookie. The problem with using cookie::Cookie is that some fields will not be forwarded to WebDriver, even when set by the user (e.g., same_site).

jonhoo: I'd actually specifically like to expose a type that isn't tied to the particular WebDriver standard so that the user doesn't have to work too hard to get to something that "just works" with the ecosystem, so let's keep this an implementation detail for the time being.

I also wonder if we should rename this `WebDriverCookie` instead, since that's what it actually represents — the fact that it's encoded as JSON is sort of neither here nor there.

jonhoo (changes requested):
This looks great, thank you! I just left one small note about the naming of the new type. I'd also prefer if the `binary` change could be in its own PR, just to try and keep the record of changes a little neater, but that's not a requirement.

src/cookies.rs
aksiksi: That makes sense!  I've renamed it to `WebDriverCookie`.

aksiksi:
@jonhoo Thank you for reviewing!  I've reverted the `binary` flag change - will open a new PR for that.  The reason I changed it in the first place is because it doesn't work on Windows (my current dev environment).

src/cookies.rs
```diff
@@ -6,12 +6,12 @@ use webdriver::command::WebDriverCommand;
 use crate::client::Client;
 use crate::error;
 
-/// Type alias for a `cookie::Cookie`
+/// Type alias for a [cookie::Cookie]
```
jonhoo, line 9: You can actually write
```
[`cookie::Cookie`]
```
and the intra-doc links feature will still pick it up!

src/cookies.rs
```diff
@@ -0,0 +1,123 @@
+//! Cookie-related functionality for WebDriver.
+use serde::{Deserialize, Serialize};
+use time::OffsetDateTime;
+use webdriver::command::WebDriverCommand;
+
+use crate::client::Client;
+use crate::error;
+
+/// Type alias for a [cookie::Cookie]
+pub type Cookie<'a> = cookie::Cookie<'a>;
```
jonhoo, line 10: Out of interest — why did you want to explicitly re-export the `Cookie` type here? Does that make a particular use-case easier?

jonhoo (approved):

jonhoo:
I'll release this once CI kicks back in :)

src/cookies.rs
aksiksi: Ah, I was thinking ahead a bit for the `add_cookie()` API!

My idea was to expose the `cookie::Cookie` that `fantoccini` pulls in so that users can construct one without pulling in the crate themselves.  In addition, this would help reduce the chance of users trying to use a more recent/non-compatible version of the `cookie` crate by mistake.

But I could also be misunderstanding how Cargo handles multiple versions of a dependency!

jonhoo: No, that seems totally reasonable! Good call :+1:

jonhoo:
I don't understand what's going on with the Travis build credits here.. Still keeping an eye on it.

jonhoo:
Finally published as 0.17.4 :tada:

aksiksi:
Awesome, thanks for following up!

PR #132, opened 2021-05-25
Remove binary flag in test client

aksiksi:
Both geckodriver and chromedriver pick the correct binary up automatically. Also, the current hardcoded path does not work on
Windows.


jonhoo (approved):
Thanks!

PR #133, opened 2021-06-03
Basic User Prompt implementation

jaboatman:
Provides an implementation for handling user prompts as defined here: https://www.w3.org/TR/webdriver1/#user-prompts

This only includes responding with `Accept` and `Dismiss`.

### DIFF
--- a/src/user_prompts.rs
+++ b/src/user_prompts.rs
@@ -12,10 +12,10 @@ use crate::Client;
 #[derive(Debug, Clone)]
 #[non_exhaustive]
 pub enum PromptAction {
-    /// `Accept` is equivalent to a user clicking the `OK` button in the prompt.
+    /// `Accept` is equivalent to a user clicking the _OK_ button in the prompt.
     Accept,
-    /// `Dismiss` is equivalent to a user clicking the `Cancel` or `OK` button in the prompt,
-    /// whichever is present and appears first.
+    /// `Dismiss` is equivalent to a user clicking the _Cancel_ or _OK_ button in the prompt,
+    /// whichever is present on the prompt, in that order.
     Dismiss,
 }
 
--- a/src/user_prompts.rs
+++ b/src/user_prompts.rs
@@ -13,7 +13,7 @@ use crate::Client;
 pub enum UserPrompt {
     /// `Accept` is equivalent to a user clicking the `OK` button in the prompt.
     Accept,
-    /// `Accept` is equivalent to a user clicking the `Cancel` or `OK` button in the prompt,
+    /// `Dismiss` is equivalent to a user clicking the `Cancel` or `OK` button in the prompt,
     /// whichever is present and appears first.
     Dismiss,
 }
--- a/src/user_prompts.rs
+++ b/src/user_prompts.rs
@@ -7,10 +7,11 @@ use webdriver::command::WebDriverCommand;
 use crate::error;
 use crate::Client;
 
-/// `UserPrompt` enumerates the different ways a `Client` can respond to a user prompt, or alert in
-/// the browser window.
+/// `PromptAction` enumerates the different actions a `Client` can take in response to a user prompt
+/// or alert in the browser window.
 #[derive(Debug, Clone)]
-pub enum UserPrompt {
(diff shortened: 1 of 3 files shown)


src/user_prompts.rs
```diff
@@ -0,0 +1,35 @@
...
+use crate::error;
+use crate::Client;
+
+/// `UserPrompt` enumerates the different ways a `Client` can respond to a user prompt, or alert in
+/// the browser window.
+#[derive(Debug, Clone)]
+pub enum UserPrompt {
+    /// `Accept` is equivalent to a user clicking the `OK` button in the prompt.
+    Accept,
+    /// `Accept` is equivalent to a user clicking the `Cancel` or `OK` button in the prompt,
```
jonhoo, line 16:
```suggestion
    /// `Dismiss` is equivalent to a user clicking the `Cancel` or `OK` button in the prompt,
```

src/user_prompts.rs
```diff
@@ -0,0 +1,35 @@
...
+//! standard.
+use webdriver::command::WebDriverCommand;
+
+use crate::error;
+use crate::Client;
+
+/// `UserPrompt` enumerates the different ways a `Client` can respond to a user prompt, or alert in
+/// the browser window.
+#[derive(Debug, Clone)]
+pub enum UserPrompt {
```
jonhoo, line 13: This should be marked `#[non_exhaustive]` so that we can add more variants (like `GetText`) later on without a breaking release.

jonhoo: Also, I wonder if this should be named `PromptAction` rather than `UserPrompt`, since it represents an action, not the prompt itself.

src/user_prompts.rs
```diff
@@ -0,0 +1,35 @@
...
+    /// `Accept` is equivalent to a user clicking the `OK` button in the prompt.
+    Accept,
+    /// `Accept` is equivalent to a user clicking the `Cancel` or `OK` button in the prompt,
+    /// whichever is present and appears first.
+    Dismiss,
+}
+
+impl Client {
+    /// Sends a response to the user prompt. For the different values you can provide, see
+    /// [UserPrompt](/UserPrompt).
```
jonhoo, line 23:
```suggestion
    /// [`UserPrompt`].
```
Intra-doc linking will take care of making this a link to the right place.

jonhoo (changes requested):
This looks great, thank you! Just left some smaller notes.

src/user_prompts.rs
jaboatman: Done.

src/user_prompts.rs
```diff
@@ -0,0 +1,39 @@
...
+
+/// `PromptAction` enumerates the different actions a `Client` can take in response to a user prompt
+/// or alert in the browser window.
+#[derive(Debug, Clone)]
+#[non_exhaustive]
+pub enum PromptAction {
+    /// `Accept` is equivalent to a user clicking the `OK` button in the prompt.
+    Accept,
+    /// `Dismiss` is equivalent to a user clicking the `Cancel` or `OK` button in the prompt,
+    /// whichever is present and appears first.
```
jonhoo, line 18: I don't think this is _quite_ accurate. The wording from the spec is:
> To dismiss the current user prompt, do so as if the user would click the Cancel or OK button, whichever is present, in that order. 

Note that it's not "which appears first", but "in that order", which is an important distinction. With the phrasing you're currently using, a user might think that if the OK button appears to the left of the Cancel button, it'd be clicked.

Also, let's remove the backticks around Cancel and OK here and above — they're not code-related. I'd go with `_` for italics instead.

jaboatman: Ha - good catch, I misunderstood the spec when reading it. I'll update it now.

jonhoo (approved):
Thanks!

jonhoo (changes requested):
Actually, would you mind adding an integration test for this too? Should be fairly simple. Just add in an HTML file that shows a prompt and sets some global value to the result, and then check the value of that variable after accepting/dismissing the prompt using the `Client::execute` method.

Issue #90, opened 2020-05-17
Can't wait_for operations other than find and navigate


Vypo:
Like I mentioned in my pull request, if elements exist in the DOM, but are hidden with CSS, operations like `click` will fail. I have a [working example][0], but the short form of my current workaround is:

```html
<!DOCTYPE html>
<html>
<body>
<a style="display: none;">foo</a>
<script>
setTimeout(() => {
let a = document.getElementsByTagName('a')[0];
a.style.display = 'block';
}, 30000);
</script>
</body>
</html>
```

```rust
let elem = client.find(Locator::Css("a")).await?;

while elem.clone().click().await.is_err() {
    std::thread::sleep(std::time::Duration::from_secs(1));
}
```

[0]: https://github.com/Vypo/fantoccini-example/blob/master/src/main.rs

jonhoo:
Thank you for the minimal example, that helps! This sort of ties into the discussion over in #86 — if `click` took `&mut self` instead of `self`, then the `clone` there wouldn't be necessary, and everything would "just work". That would also make it possible to write this using `wait_for`, since `&mut Client` would then (I believe) be sufficient.

PR #134, opened 2021-06-21
Implement "wait for" functions with a timeout

ctron:
This is first proposal for the following issues:

Fixes https://github.com/jonhoo/fantoccini/issues/81
Fixes https://github.com/jonhoo/fantoccini/issues/90
Fixes https://github.com/jonhoo/fantoccini/issues/27

I tried to:

* Avoid cloning the client
* Allow for future extensions on the wait parameters
* Provide a single non-opinionated "on" method
* Provide more opinionated function which make your live easier (on_element, on_predicate)

I didn't not async-trait, as it wasn't in the project already. Maybe that could help creating a better API.

One thing I wasn't able to achieve so far, is to provide a way directly pass in e.g. a `Locator` into the `on` method, and let it `Into` an appropriate function. I think that would be nice, but it also is a bit above my Rust future and lifetime knowledge.

It currently also lacks proper documentation.

### DIFF
--- a/examples/basic.rs
+++ b/examples/basic.rs
@@ -39,7 +39,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     // Click the "Get Started" button.
     let button = client
         .wait()
-        .on_element(Locator::Css(
+        .on(Locator::Css(
             r#"a.button-download[href="/learn/get-started"]"#,
         ))
         .await?;
@@ -47,21 +47,18 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     // Click the "Try Rust Without Installing" button (using XPath this time).
     let button = r#"//a[@class="button button-secondary" and @href="https://play.rust-lang.org/"]"#;
-    let button = client.wait().on_element(Locator::XPath(button)).await?;
+    let button = client.wait().on(Locator::XPath(button)).await?;
     button.click().await?;
 
     // Find the big textarea.
-    let mut code_area = client
-        .wait()
-        .on_element(Locator::Css(".ace_text-input"))
-        .await?;
+    let mut code_area = client.wait().on(Locator::Css(".ace_text-input")).await?;
 
     // And write in some code.
     code_area.send_keys("// Hello from Fantoccini\n").await?;
 
     // Now, let's run it!
     let button = r#"//div[@class="segmented-button"]/button[1]"#;
-    let button = client.wait().on_element(Locator::XPath(button)).await?;
+    let button = client.wait().on(Locator::XPath(button)).await?;
     button.click().await?;
 
     // Let the user marvel at what we achieved.
--- a/examples/basic.rs
+++ b/examples/basic.rs
@@ -36,34 +36,32 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     // Go to the Rust website.
     client.goto("https://www.rust-lang.org/").await?;
 
-    // This sleep is just used to make the browser's actions visible.
-    sleep(Duration::from_millis(1000)).await;
-
     // Click the "Get Started" button.
     let button = client
-        .find(Locator::Css(
+        .wait()
+        .on_element(Locator::Css(
             r#"a.button-download[href="/learn/get-started"]"#,
         ))
         .await?;
     button.click().await?;
-    sleep(Duration::from_millis(1000)).await;
 
     // Click the "Try Rust Without Installing" button (using XPath this time).
     let button = r#"//a[@class="button button-secondary" and @href="https://play.rust-lang.org/"]"#;
-    let button = client.find(Locator::XPath(button)).await?;
+    let button = client.wait().on_element(Locator::XPath(button)).await?;
     button.click().await?;
-    sleep(Duration::from_millis(1000)).await;
 
     // Find the big textarea.
-    let mut code_area = client.find(Locator::Css(".ace_text-input")).await?;
+    let mut code_area = client
+        .wait()
+        .on_element(Locator::Css(".ace_text-input"))
+        .await?;
 
     // And write in some code.
     code_area.send_keys("// Hello from Fantoccini\n").await?;
-    sleep(Duration::from_millis(1000)).await;
 
     // Now, let's run it!
     let button = r#"//div[@class="segmented-button"]/button[1]"#;
-    let button = client.find(Locator::XPath(button)).await?;
+    let button = client.wait().on_element(Locator::XPath(button)).await?;
     button.click().await?;
 
     // Let the user marvel at what we achieved.
--- a/src/client.rs
+++ b/src/client.rs
@@ -671,13 +671,7 @@ impl Client {
         note = "This method might block forever. Please use client.wait().on(locator) instead. You can still wait forever using: client.wait().forever().on(locator)"
     )]
     pub async fn wait_for_find(&mut self, search: Locator<'_>) -> Result<Element, error::CmdError> {
-        loop {
-            match self.wait().forever().on(search).await {
-                Ok(ele) => break Ok(ele),
-                Err(error::CmdError::WaitTimeout) => continue,
-                Err(err) => break Err(err),
-            }
-        }
+        self.wait().forever().on(search).await
     }
 
     /// Wait for the page to navigate to a new URL before proceeding.
--- a/src/client.rs
+++ b/src/client.rs
@@ -647,6 +647,10 @@ impl Client {
     /// While this currently just spins and yields, it may be more efficient than this in the
     /// future. In particular, in time, it may only run `is_ready` again when an event occurs on
     /// the page.
+    #[deprecated(
+        since = "0.17.5",
+        note = "This method might block forever. Please use client.wait().on(...) instead. You can still wait forever using: client.wait().forever().on(...)"
+    )]
     pub async fn wait_for<F, FF>(&mut self, mut is_ready: F) -> Result<(), error::CmdError>
     where
         F: FnMut(&mut Client) -> FF,
@@ -662,19 +666,16 @@ impl Client {
     /// While this currently just spins and yields, it may be more efficient than this in the
     /// future. In particular, in time, it may only run `is_ready` again when an event occurs on
     /// the page.
+    #[deprecated(
+        since = "0.17.5",
+        note = "This method might block forever. Please use client.wait().on(locator) instead. You can still wait forever using: client.wait().forever().on(locator)"
+    )]
     pub async fn wait_for_find(&mut self, search: Locator<'_>) -> Result<Element, error::CmdError> {
-        let s: webdriver::command::LocatorParameters = search.into();
         loop {
-            match self
-                .by(webdriver::command::LocatorParameters {
-                    using: s.using,
-                    value: s.value.clone(),
-                })
-                .await
-            {
-                Ok(v) => break Ok(v),
-                Err(error::CmdError::NoSuchElement(_)) => {}
-                Err(e) => break Err(e),
+            match self.wait().forever().on(search).await {
+                Ok(ele) => break Ok(ele),
+                Err(error::CmdError::WaitTimeout) => continue,
+                Err(err) => break Err(err),
             }
         }
     }
@@ -684,6 +685,10 @@ impl Client {
     /// If the `current` URL is not provided, `self.current_url()` will be used. Note however that
     /// this introduces a race condition: the browser could finish navigating *before* we call
     /// `current_url()`, which would lead to an eternal wait.
+    #[deprecated(
+        since = "0.17.5",
+        note = "This method might block forever and has a chance of randomly blocking. Please use client.wait().on(url) instead, to check if you navigated successfully to the new URL."
+    )]
     pub async fn wait_for_navigation(
         &mut self,
         current: Option<url::Url>,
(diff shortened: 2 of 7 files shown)


ctron:
@jonhoo Do you have any feedback on this PR?

jonhoo:
@ctron Sorry — this is on my radar, but I just haven't had the spare cycles to give it a proper review! I'm hoping I'll get to it next week.

ctron:
> @ctron Sorry — this is on my radar, but I just haven't had the spare cycles to give it a proper review! I'm hoping I'll get to it next week.

No worries, thanks for letting me know!

src/wait.rs
```diff
@@ -0,0 +1,155 @@
...
+    pub fn new(client: &'c mut Client) -> Self {
+        Self {
+            client,
+            timeout: Some(DEFAULT_TIMEOUT),
+            period: DEFAULT_PERIOD,
+        }
+    }
+
+    /// Set the timeout until the operation should wait.
+    pub fn until(mut self, timeout: Duration) -> Self {
```
jonhoo, line 33: How about the name `at_most` instead? `until` implies it takes an `Instant`/`SystemTime` to me at least, not a `Duration`.

src/wait.rs
```diff
@@ -0,0 +1,155 @@
...
+        self
+    }
+
+    /// Wait until a condition exists or a timeout is hit.
+    pub async fn on<T, F>(self, mut f: F) -> Result<T, WaitError>
+    where
+        F: for<'f> FnMut(
+            &'f mut Client,
+        )
+            -> Pin<Box<dyn Future<Output = Result<Option<T>, error::CmdError>> + 'f>>,
```
jonhoo, line 56: I don't think you need `Pin<Box<dyn` here. Does this work?
```suggestion
    pub async fn on<T, F, Fut>(self, mut f: F) -> Result<T, WaitError>
    where
        F: for<'f> FnMut(&'f mut Client) -> Fut + 'f,
        Fut: Future<Output = Result<Option<T>, error::CmdError>
```

jonhoo: In fact, you may even be able to remove all mention of `'f`

src/wait.rs
```diff
@@ -0,0 +1,155 @@
...
+}
+
+/// Allow creating a wait operation.
+pub trait CanWait {
+    /// Create a new wait operation.
+    fn wait(&mut self) -> Wait<'_>;
+}
+
+impl CanWait for Client {
+    fn wait(&mut self) -> Wait<'_> {
```
jonhoo, line 112: I would just make this a method directly on `Client` rather than introducing a trait just for this for the one implementation.

jonhoo (changes requested):
Finally got around to this! This looks great, thank you. I left some notes inline — I think we can avoid the `Pin<Box<dyn Future>>` since we're not actually in trait context here at all.

I'd also like to see some documentation and tests for this :+1:

src/wait.rs
ctron: Yes it works :grinning: 

Dropping either the `'f` lifetime, or using a extra type argument for the future (so that I don't need to use a PinBox) both makes me run into lifetime issue, as the compiler can't figure it out on its own.

The goal is to use the client inside the "on" function. So the function must be valid for the lifetime of the client only, but not necessarily for `'static'`, what the compiler falls back to (complaining that the function is not valid for `'static`).

Using a [Higher-Rank Trait Bound](https://doc.rust-lang.org/nomicon/hrtb.html#higher-rank-trait-bounds-hrtbs) seems to do the trick.

Having said that, maybe there is a better way do deal with this. I would be happy to learn about it :grin:.

src/wait.rs
ctron: Sure. I wanted to keep it out client for the moment, to not clash with any existing functionality.

ctron:
I merged the "CanWait" trait directly into the client. I also updated the example to use the new functions.

I was thinking how to treat the current `wait_*` functions on the `Client`. One way to deal with this could be to mark them as deprecated and re-implement them with the new functionality, using `forever` in the process to keep the existing behavior.

jonhoo:
> One way to deal with this could be to mark them as deprecated and re-implement them with the new functionality, using `forever` in the process to keep the existing behavior.

I like that!

src/wait.rs
jonhoo: Interesting, that's unfortunate. Well, let's keep in `PinBox` then :sweat_smile:

ctron:
Sorry for the long delay. I finally found some more time looking into this.

While trying to replace the existing methods, I was wondering if I could have a more generic approach to the `on_*` wait methods. And came up with a `trait`. What you can do now is to directly drop in stuff to wait for:

~~~rust
let _: Element = client
    .wait()
    .on(Locator::Css(
        r#"a.button-download[href="/learn/get-started"]"#,
    ))
    .await?;
~~~

I hope you like it, but I can also rip it out again if you don't. With that, I was also able to drop the `'static`  requirement on the Locator variant.

One thing I noticed when switching the tests to use the new functions, and re-implementing the `wait_*` functions was, that the `WaitError` was a nuisance. You have mostly `CmdError` and once methods want to wait, you need to use a generic error type, or deal with the timeout explicitly. That felt annoying. So I dropped the error and moved it into the `CmdError`. You can still check for the timeout condition, but it is a "normal" error now. Again, if you don't like it, I can take it out again. But as the wait functionality is in the client now, I assumed that it is ok.

I failed a bit with my original goal, of replacing the `wait_*` functions. I marked the deprecated, giving some guidance of how to deal with that. But the Fn definition on the `wait_for` method was a bit broader than the new one. So I could do all kind of tricks to work around this, or the leave the existing two line implementation in place … :shrug:  I was able to replace one of them. The third one was based on `wait_for` anyway, and felt a bit broken. Let me know what you think. I can give it another try if you like.

src/wait.rs
```diff
@@ -0,0 +1,288 @@
...
+    {
+        self.on(Predicate(predicate)).await
+    }
+}
+
+/// A condition to wait for.
+///
+/// This is implemented by different types directly, like [`Locator`]. But you can also use
+/// custom logic, using the [`Condition`] and [`Predicate`] newtypes.
+pub trait WaitCondition<'a, T> {
```
jonhoo, line 176: I don't think you need `'a` here at all actually. Just make `ready` generic over `<'a>` instead. That will also simplify your bounds in the methods above further.

src/wait.rs
```diff
@@ -0,0 +1,288 @@
...
+//! # Error handling
+//!
+//! When a wait operation times out, it will return a [`CmdError::WaitTimeout`]. When a wait
+//! condition check returns an error, the wait operation will be aborted, and the error returned.
+//!
+//! # Custom conditions
+//!
+//! You can implement custom conditions either by implementing the [`WaitCondition`] trait or by
+//! using closures. Due to lifetime and async trait difficulties, two newtypes and two dedicated
+//! functions exists to simplify the usage of closures. Also see: [`Closure`], [`Predicate`].
```
jonhoo, line 49:
```suggestion
//! using closures. If you just want a closure, use [`Wait::until`] and [`Wait::until_some`].
```

I don't think anyone will want to use `Closure` and `Predicate` directly.

src/wait.rs
```diff
@@ -0,0 +1,288 @@
...
+}
+
+/// Newtype for condition functions.
+///
+/// A condition is successful once it returns some value. It will be re-tried as long as it
+/// returns [`None`].
+///
+/// Instead of using the newtype, you can also use the [`Wait::on_condition`] method.
+#[derive(Debug)]
+pub struct Condition<F, T>(pub F)
```
jonhoo, line 234: Is there a particular reason you want this type to be public? I'd just make it private. Same with `Predicate`.

jonhoo (changes requested):
I like the new direction! Left some inline comments.

examples/basic.rs
```diff
@@ -36,34 +36,29 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
     client.goto("https://www.rust-lang.org/").await?;
 
-    // This sleep is just used to make the browser's actions visible.
-    sleep(Duration::from_millis(1000)).await;
-
     // Click the "Get Started" button.
     let button = client
-        .find(Locator::Css(
+        .wait()
+        .on(Locator::Css(
```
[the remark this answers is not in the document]
ctron: You are waiting some something to be located … technically, a `WaitCondition` returns a value (and that may also be `()`). I like the compactness of `on`. You are waiting "on" ("for" would be keyword) something.

Creating an `on_find` method would be different than the current `on` method. Just that `on` would need to discard the return value.

Not sure that is improving the API though.

src/client.rs
```diff
@@ -661,19 +666,16 @@ impl Client {
...
-                .by(webdriver::command::LocatorParameters {
-                    using: s.using,
-                    value: s.value.clone(),
-                })
-                .await
-            {
-                Ok(v) => break Ok(v),
-                Err(error::CmdError::NoSuchElement(_)) => {}
-                Err(e) => break Err(e),
+            match self.wait().forever().on(search).await {
```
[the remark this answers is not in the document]
ctron: Indeed no :)

src/wait.rs
ctron: Yes, I wanted to allow this to be used a parameter to `on`. This way one can keep the pattern or just using `on` for everything: Locator, Url, …

src/wait.rs
```diff
@@ -0,0 +1,288 @@
...
+                Some(result) => break Ok(result),
+                None => {
+                    tokio::time::sleep(self.period).await;
+                }
+            };
+        }
+    }
+
+    /// Wait for a predicate.
+    pub async fn on_predicate<F>(self, predicate: F) -> Result<(), error::CmdError>
```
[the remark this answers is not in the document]
ctron: I don't think "while" would work, as it is a keyword and using `r#while` would seem a bit odd. I will go for `until` then.

ctron:
Glad you liked it. I updated the PR to incorporate your feedback.  I am just not so sure about the return type comment.

Regarding the lifetime changes, I wasn't able to apply the proposed changes. To my understanding, the lifetimes of the client, the reference to the client, and the closures are different. And that is the only way I was able to come up with that works. And I am not saying that I understand all of it :)

Maybe go ahead and play with it yourself. I am happy to learn a few new tricks when it comes to lifetimes :)

src/wait.rs
jonhoo: I think we should make it private — in practice I don't think there's that much value in supporting people passing lots of different types to `on` directly.

examples/basic.rs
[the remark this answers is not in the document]
jonhoo: Yeah, that's fair. It just reads a little oddly, but I suppose `on_find` wouldn't _really_ be all that much nicer.

src/wait.rs
jonhoo: Did you try making this change? I think that alone will simplify a lot of your bounds.

ctron: I tried it before, and I failed with the blanket implementation of `WaitCondition` for the Fn type.

However, that isn't really being used, as I switched to the `Condition` and `Predicate` newtypes anyway, one more reason why I would like to keep them.

So `'a` is dropped now, and it still works with the blanket implementation dropped.

src/wait.rs
ctron: I personally prefer a generic `on` method over the specialized `on_predicate` methods. If you really don't like these types, maybe we can just hide them from the documentation, and let people use them when they want.

I don't see much benefit in hiding this. I think both ways (newtype and specialized method) are valid approaches, and would simply let the user decide.

jonhoo: So, there are two reasons I don't like the generic `on`:

1. It means that if, in the future, we get a more reliable way to wait for an element's appearance than just spinning with a delay, we won't be able to take advantage of it easily. For example, for locators, you could imagine using the observer API to make the browser notify us when a locator matches, rather than just retrying after a certain amount of time. Retrofitting that into `on` is tricky when it's generic, since we'd only be able to do it _specifically_ for `Locator`.
2. It means that there are multiple ways to do the same thing. That makes the API more complicated for users, and means we need to support twice as many use-cases. If we instead _just_ have dedicated methods, it's very obvious to users how they do each of the waiting operations, and they don't have to worry about the trait bounds or the `WaitCondition` trait.

So, I'm very much partial to just stripping this down to `fn on(Locator)`, `fn until(impl Fn(&mut Client) -> bool)`, and `fn until_some(impl Fn(&mut Client) -> Option<T>)`. Honestly, I would probably skip the `-> bool` version too, but could be convinced otherwise.

jonhoo:
Glad you managed to simplify the lifetime things! :tada:

ctron:
I was away for a few days. Is there something you want to me to fix up. Could you further improve the lifetimes?

jonhoo:
I'm going to go ahead and merge this, and then fiddle a little more on my own. Thanks for all your work on this, and for sticking with it!

ctron:
Thanks for merging. And thanks for your feedback! I think this improved the result quite a lot!

And: I am looking forward to a new release now :grinning:

jonhoo:
I actually ended up removing the generic interface, and stuck with just the concrete methods for `Url` and `Locator`. My thinking is that for those we may be able to provide smarter retries over time (e.g., using JavaScript event observers). The same isn't true for general-purpose asynchronous closures, so I'd rather not give users that impression. It's pretty easy to implement one's own loop (it's really just a `loop` with a `tokio::time::sleep`), so I'd rather keep that in the caller's code.

PR #139, opened 2021-10-28
Upgrade outdated dependencies and track them with dependabot

tyranron:
I've bumped crate version to `0.18.0`, as `webdriver` upgraded from `0.43` to `0.44`, so it's a breaking change.

Also, I've used `0.16.0-rc.1` for cookie as it uses `0.3` version of `time`. This doesn't count in public APIs, so transitioning to `0.16` should be non-breaking in future.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: cargo
+    directory: /
+    schedule:
+      interval: daily
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,6 +1,0 @@
-version: 2
-updates:
-  - package-ecosystem: cargo
-    directory: /
-    schedule:
-      interval: daily
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,7 +40,7 @@ http = "0.2"
 time = "0.3"
 
 [dev-dependencies]
-tokio = { version = "1.0", features = ["full"] }
+tokio = { version = "1", features = ["full"] }
 hyper = { version = "0.14", features = ["server", "tcp"] }
 serial_test = "0.5"
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,8 +33,8 @@ tokio = { version = "1.0", features = ["sync", "rt"] }
 hyper = { version = "0.14", features = ["stream", "client", "http1", "http2"] }
 cookie = { version = "0.16.0-rc.1", features = ["percent-encode"] }
 base64 = "0.13"
-hyper-rustls = { version = "0.23", optional = true }
-hyper-tls = { version = "0.5", optional = true }
+hyper-rustls = { version = "0.23.0", optional = true }
+hyper-tls = { version = "0.5.0", optional = true }
 mime = "0.3.9"
 http = "0.2"
 time = "0.3"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "fantoccini"
-version = "0.17.6"
+version = "0.18.0"
 edition = "2018"
 
 description = "High-level API for programmatically interacting with web pages through WebDriver."
@@ -23,25 +23,25 @@ native-tls = ["hyper-tls"]
 rustls-tls = ["hyper-rustls"]
 
 [dependencies]
-webdriver = { version = "0.43.0", default-features = false }
+webdriver = { version = "0.44", default-features = false }
 url = "2.2.2"
-serde = { version = "1.0", features = [ "derive" ] }
+serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
-futures-core = "0.3.0"
-futures-util = "0.3.0"
-tokio = { version = "1", features = [ "sync", "rt" ] }
-hyper = { version = "0.14", features = [ "stream", "client", "http1", "http2" ] }
-cookie = { version = "0.14", features = ["percent-encode"] }
+futures-core = "0.3"
+futures-util = "0.3"
+tokio = { version = "1.0", features = ["sync", "rt"] }
+hyper = { version = "0.14", features = ["stream", "client", "http1", "http2"] }
+cookie = { version = "0.16.0-rc.1", features = ["percent-encode"] }
 base64 = "0.13"
 hyper-rustls = { version = "0.22.1", optional = true }
-hyper-tls = { version = "0.5.0", optional = true }
+hyper-tls = { version = "0.5", optional = true }
 mime = "0.3.9"
 http = "0.2"
-time = "0.2.26"
+time = "0.3"
 
 [dev-dependencies]
-tokio = { version = "1", features = [ "full" ] }
-hyper = { version = "0.14", features = [ "server", "tcp" ] }
+tokio = { version = "1.0", features = ["full"] }
+hyper = { version = "0.14", features = ["server", "tcp"] }
 serial_test = "0.5"
 
 [package.metadata.docs.rs]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,7 @@ tokio = { version = "1.0", features = ["sync", "rt"] }
 hyper = { version = "0.14", features = ["stream", "client", "http1", "http2"] }
 cookie = { version = "0.16.0-rc.1", features = ["percent-encode"] }
 base64 = "0.13"
-hyper-rustls = { version = "0.22.1", optional = true }
+hyper-rustls = { version = "0.23", optional = true }
 hyper-tls = { version = "0.5", optional = true }
 mime = "0.3.9"
 http = "0.2"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 futures-core = "0.3"
 futures-util = "0.3"
-tokio = { version = "1.0", features = ["sync", "rt"] }
+tokio = { version = "1", features = ["sync", "rt"] }
 hyper = { version = "0.14", features = ["stream", "client", "http1", "http2"] }
 cookie = { version = "0.16.0-rc.1", features = ["percent-encode"] }
 base64 = "0.13"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "fantoccini"
-version = "0.18.0"
+version = "0.17.6"
 edition = "2018"
 
 description = "High-level API for programmatically interacting with web pages through WebDriver."
--- a/src/client.rs
+++ b/src/client.rs
@@ -17,7 +17,7 @@ use webdriver::common::{FrameId, ELEMENT_KEY};
 use crate::ClientBuilder;
 
 /// A WebDriver client tied to a single browser
-/// [session](https://www.w3.org/TR/webdriver1/#sessions).
+/// [session](https://w3.org/TR/webdriver1/#sessions).
 ///
 /// Use [`ClientBuilder`](crate::ClientBuilder) to create a new session.
 ///
@@ -145,11 +145,11 @@ impl Client {
 
 // NOTE: new impl block to keep related methods together.
 
-/// [Navigation](https://www.w3.org/TR/webdriver1/#navigation)
+/// [Navigation](https://w3.org/TR/webdriver1/#navigation)
 impl Client {
     /// Navigate directly to the given URL.
     ///
-    /// See [9.1 Navigate To](https://www.w3.org/TR/webdriver1/#dfn-navigate-to) of the WebDriver
+    /// See [9.1 Navigate To](https://w3.org/TR/webdriver1/#dfn-navigate-to) of the WebDriver
     /// standard.
     #[cfg_attr(docsrs, doc(alias = "Navigate To"))]
     pub async fn goto(&mut self, url: &str) -> Result<(), error::CmdError> {
@@ -165,7 +165,7 @@ impl Client {
 
     /// Retrieve the currently active URL for this session.
     ///
-    /// See [9.2 Get Current URL](https://www.w3.org/TR/webdriver1/#dfn-get-current-url) of the
+    /// See [9.2 Get Current URL](https://w3.org/TR/webdriver1/#dfn-get-current-url) of the
     /// WebDriver standard.
     #[cfg_attr(docsrs, doc(alias = "Get Current URL"))]
     pub async fn current_url(&mut self) -> Result<url::Url, error::CmdError> {
@@ -184,7 +184,7 @@ impl Client {
 
     /// Go back to the previous page.
     ///
-    /// See [9.3 Back](https://www.w3.org/TR/webdriver1/#dfn-back) of the WebDriver standard.
+    /// See [9.3 Back](https://w3.org/TR/webdriver1/#dfn-back) of the WebDriver standard.
     #[cfg_attr(docsrs, doc(alias = "Back"))]
     pub async fn back(&mut self) -> Result<(), error::CmdError> {
         self.issue(WebDriverCommand::GoBack).await?;
@@ -193,7 +193,7 @@ impl Client {
 
     /// Refresh the current previous page.
     ///
-    /// See [9.5 Refresh](https://www.w3.org/TR/webdriver1/#dfn-refresh) of the WebDriver standard.
+    /// See [9.5 Refresh](https://w3.org/TR/webdriver1/#dfn-refresh) of the WebDriver standard.
     #[cfg_attr(docsrs, doc(alias = "Refresh"))]
     pub async fn refresh(&mut self) -> Result<(), error::CmdError> {
         self.issue(WebDriverCommand::Refresh).await?;
@@ -201,11 +201,11 @@ impl Client {
     }
 }
 
-/// [Command Contexts](https://www.w3.org/TR/webdriver1/#command-contexts)
+/// [Command Contexts](https://w3.org/TR/webdriver1/#command-contexts)
 impl Client {
     /// Gets the current window handle.
     ///
-    /// See [10.1 Get Window Handle](https://www.w3.org/TR/webdriver1/#get-window-handle) of the
+    /// See [10.1 Get Window Handle](https://w3.org/TR/webdriver1/#get-window-handle) of the
     /// WebDriver standard.
     #[cfg_attr(docsrs, doc(alias = "Get Window Handle"))]
     pub async fn window(&mut self) -> Result<WindowHandle, error::CmdError> {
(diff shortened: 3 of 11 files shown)


jonhoo:
The big reason not to do a breaking release, even if it requires _basically_ no action on the part of consumers, is that it still requires them to _realize_ that the major version has changed and go update their `Cargo.toml`. If they don't, they'll continue being on an increasingly outdated version. Whereas if we stay on the same major version, they don't need to take any action to keep getting updates.

I'm actually not so much worried about dependents having taken dependencies on `webdriver` directly — I'd honestly be surprised if they have. What I'd _really_ like to see is for us to change fantoccini so that it doesn't expose _any_ types from `webdriver` (we really want it to be a [private dependency](https://github.com/rust-lang/rust/issues/44663)). That way we'd be able to bump webdriver in the future without doing breaking releases. Now _that's_ a change I'd be willing to cut a breaking release for :p

Any chance you'd be willing to take on the work to get rid of `webdriver` types that appear in the public API?

src/lib.rs
```diff
@@ -266,29 +270,79 @@ pub enum Locator<'a> {
...
 }
 
-mod client;
+/// A [handle][1] to a browser window.
+///
+/// Should be obtained it via [`Client::window()`] method (or similar).
+///
+/// [1]: https://w3.org/TR/webdriver/#dfn-window-handles
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct WindowHandle(String);
```
jonhoo, line 301: We should probably split these newtypes into their own file now that we're like to get more of them

tyranron:
@jonhoo well, I've ripped out `webdriver` types usages in public APIs, there weren't much:
- `webdriver::common::WebWindow` -> `WindowHandle` opaque tuple struct
- `webdriver::response::NewWindowResponse` -> `NewWindowResponse` struct + `NewWindowType` enum
- `webdriver::capabilities::Capabilities` -> `Capabilities` alias (actually, the same thing, but without referring upstream)

That was an "easy-peasy" part.

The harder part, where I need your guidance is whether we should to rip out `webdriver::error::WebDriverError` from our `error::*` types?   
It's quite an amount of effort not only for implementing, but also for supporting that in future (a lot of duplication). But without doing it, could we really say that we won't introduce BC break on `webdriver`'s version bump? Seems that for `Display`ing it's OK, but if someone has explicitly `match`ed over `WebDriverError` that won't work.

src/lib.rs
tyranron: @jonhoo which module name would you prefer? I'm unsure here.

jonhoo:
Hmm, for errors I'm tempted to type-erase them to `Box<dyn Error>`, and then users can downcast if they truly need to get at the underlying WebDriver error. The downside of that of course is if matching on the exact error variant is common, but I don't have a great sense for how likely that is :thinking: What do you think?

src/lib.rs
jonhoo: `mod w3c` or `mod webdriver` maybe?

tyranron:
@jonhoo I don't think the matching on `webdriver` errors is common. We use `fantoccini` quite long and extensively in our E2E tests for doing WebRTC stuff. And we never had a case to match over errors. Just printing them was always the case.

> I'm tempted to type-erase them to `Box<dyn Error>`, and then users can downcast if they truly need to get at the underlying WebDriver error.

I somewhat doubt that solves the problem. From some formal point of view - yes, definitely. But for the actual code that may match on `webdriver::Error` we don't do any better. The situations is the same, the implementor still will do need to manage its direct `webdriver` dependency. The only thing that has changed - we're breaking his code somewhat more silently 🙈 (instead of compilation error he will received a branch misdirection in runtime).

So, as I see, either we live `webdriver` in errors and don't bother about that.  Or we erase them completely without an ability to downcast into the original one.

jonhoo:
That's true, but at least it strongly discourages matching on `webdriver::Error`. Your point is a good one though, it's not a great solution. How about this (painful) solution: we make our own `enum error::WebDriver` type that we a) mark as `#[non_exhaustive]` and b) manually add all the variants that exist in `webdriver::WebDriver` to. We also add a private helper method to convert any `webdriver::Error` into `error::WebDriver`. Then, we only use the latter type in our APIs. It's more maintenance work, but it _should_ disconnect us from `webdriver` versioning.

tyranron:
@jonhoo finally, found some time to finish this.

The "error ripping" stuff was easier than I though it would be. So now we have `error::WebDriver` which does't use upstream `WebDriverError` in public APIs, while allows library users to match over error codes by their [spec values](https://www.w3.org/TR/webdriver/#dfn-error-code) (represented as `&'static str`) in a fully backwards-compatible way.

Also, ripped out any `webdriver` type usages in `From` conversions, and replaced them with custom `pub(crate)` functions, so the  `webdriver` types won't appear in traits too.

With all this we can say now that `fantoccini` public API is fully indepedent on `webdriver` crate. 🎉

Addiionally:
- Moved new definitions (along with the `Locator`) to `wd` module (to not be confused with `webdriver` crate).
- Replaced `.map(|_| ())` pattern over the code with `.map(drop)`.
- Shortened `www.w3.org` links to just `w3.org`.

The CI actually passes, the [`beta` build failed](https://app.travis-ci.com/github/jonhoo/fantoccini/jobs/549332817) because of:
```
The job exceeded the maximum log length, and has been terminated.
```
While `stable` and `nightly` [are OK](https://github.com/jonhoo/fantoccini/pull/139/checks?check_run_id=4300631589).

Final commit message for squash merge:
```
Remove `webdriver` crate  types from public APIs (#139)

- replace `webdriver::common::WebWindow` with `wd::WindowHandle`
- replace `webdriver::response::NewWindowResponse` -> `client::NewWindowResponse` struct + `wd::NewWindowType` enum
- replace `webdriver::capabilities::Capabilities` with `wd::Capabilities`
- replace `webdriver::error::WebDriverError` with `error::WebDriver`

Additionally:
- upgrade outdated Cargo dependencies
- enable dependabot to track outdated dependencies
- replace `.map(|_| ())` code pattern with `.map(drop)`
- shorten `www.w3.org` links to just `w3.org`
```

tyranron:
ping @jonhoo

tyranron:
ping @jonhoo

.github/dependabot.yml
```diff
@@ -0,0 +1,6 @@
+version: 2
```
jonhoo, line 1: Let's do this in a separate PR.

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "fantoccini"
-version = "0.17.6"
+version = "0.18.0"
```
jonhoo, line 3: I like to do these version bumps separately rather than in-PR.

Cargo.toml
```diff
@@ -23,25 +23,25 @@ native-tls = ["hyper-tls"]
...
+serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
-futures-core = "0.3.0"
-futures-util = "0.3.0"
-tokio = { version = "1", features = [ "sync", "rt" ] }
-hyper = { version = "0.14", features = [ "stream", "client", "http1", "http2" ] }
-cookie = { version = "0.14", features = ["percent-encode"] }
+futures-core = "0.3"
+futures-util = "0.3"
+tokio = { version = "1.0", features = ["sync", "rt"] }
```
jonhoo, line 32:
```suggestion
tokio = { version = "1", features = ["sync", "rt"] }
```
Adding the .0 doesn't do anything meaningful here.

Cargo.toml
```diff
@@ -23,25 +23,25 @@ native-tls = ["hyper-tls"]
...
+hyper = { version = "0.14", features = ["stream", "client", "http1", "http2"] }
+cookie = { version = "0.16.0-rc.1", features = ["percent-encode"] }
```
jonhoo, line 34: This is a very unfortunate dependency given that we expose types from `cookie` in the public API, and so when we switch to the non-rc version, that'll arguably be a breaking change. I may end up releasing this change as an `rc` as well until `cookie` has a full release.

Cargo.toml
```diff
@@ -23,25 +23,25 @@ native-tls = ["hyper-tls"]
...
 hyper-tls = { version = "0.5.0", optional = true }
 mime = "0.3.9"
 http = "0.2"
-time = "0.2.26"
+time = "0.3"
 
 [dev-dependencies]
-tokio = { version = "1", features = [ "full" ] }
-hyper = { version = "0.14", features = [ "server", "tcp" ] }
+tokio = { version = "1.0", features = ["full"] }
```
jonhoo, line 43:
```suggestion
tokio = { version = "1", features = ["full"] }
```

src/cookies.rs
```diff
@@ -9,7 +9,7 @@ use crate::error;
 /// Type alias for a [cookie::Cookie]
 pub type Cookie<'a> = cookie::Cookie<'a>;
 
-/// Representation of a cookie as [defined by WebDriver](https://www.w3.org/TR/webdriver1/#cookies).
+/// Representation of a cookie as [defined by WebDriver](https://w3.org/TR/webdriver1/#cookies).
```
jonhoo, line 12: Why remove all the `www`s? That's the canonical URL that ends up being redirected to to after all?

src/cookies.rs
```diff
@@ -108,21 +110,19 @@ impl Client {
...
     /// Delete a single cookie from the current document.
     ///
-    /// See [16.4 Delete Cookie](https://www.w3.org/TR/webdriver1/#delete-cookie) of the
+    /// See [16.4 Delete Cookie](https://w3.org/TR/webdriver1/#delete-cookie) of the
     /// WebDriver standard.
     pub async fn delete_cookie(&mut self, name: &str) -> Result<(), error::CmdError> {
         self.issue(WebDriverCommand::DeleteCookie(name.to_string()))
             .await
-            .map(|_| ())
+            .map(drop)
```
jonhoo, line 118:
```suggestion
            .map(|_| ())
```
I don't think `map(drop)` is meaningfully clearer. Did clippy complain?

src/cookies.rs
```diff
@@ -108,21 +110,19 @@ impl Client {
...
     /// Delete all cookies from the current document.
     ///
-    /// See [16.5 Delete All Cookies](https://www.w3.org/TR/webdriver1/#delete-all-cookies) of the
+    /// See [16.5 Delete All Cookies](https://w3.org/TR/webdriver1/#delete-all-cookies) of the
     /// WebDriver standard.
     pub async fn delete_all_cookies(&mut self) -> Result<(), error::CmdError> {
-        self.issue(WebDriverCommand::DeleteCookies)
-            .await
-            .map(|_| ())
+        self.issue(WebDriverCommand::DeleteCookies).await.map(drop)
```
jonhoo, line 126: I'd like to undo this change as well unless clippy insists.

src/elements.rs
```diff
@@ -1,17 +1,18 @@
 //! Types used to represent particular elements on a page.
 
-use crate::{error, Client, Locator};
+use crate::wd::Locator;
+use crate::{error, Client};
 use serde::Serialize;
 use serde_json::Value as Json;
-use webdriver::command::{SendKeysParameters, SwitchToFrameParameters, WebDriverCommand};
```
jonhoo, line 6: Why remove these `use`s?

src/lib.rs
```diff
@@ -248,56 +251,15 @@ where
...
-            Locator::LinkText(s) => webdriver::command::LocatorParameters {
-                using: webdriver::common::LocatorStrategy::LinkText,
-                value: s.to_string(),
-            },
-        }
-    }
-}
-
-mod client;
+pub mod client;
```
jonhoo, line 254: Why make this module public?

src/wd.rs
```diff
@@ -0,0 +1,152 @@
...
+    ///
+    /// Avoids allocation if possible.
+    ///
+    /// # Errors
+    ///
+    /// If the given string is [`"current"`][1].
+    ///
+    /// [1]: https://w3.org/TR/webdriver/#dfn-window-handles
+    fn try_from(s: Cow<'a, str>) -> Result<Self, Self::Error> {
+        if s != "current" {
```
jonhoo, line 35: I'm not sure I follow why this is disallowed for _all_ instances of `WindowHandle`? I thought it was permitted as an argument for selecting a window?

src/client.rs
```diff
@@ -1,23 +1,24 @@
...
 use crate::wait::Wait;
-use crate::{error, Locator};
+use crate::wd::{Capabilities, Locator, NewWindowType, WindowHandle};
 use hyper::{client::connect, Method};
 use serde_json::Value as Json;
-use std::convert::TryFrom;
+use std::convert::{TryFrom, TryInto as _};
 use std::future::Future;
 use tokio::sync::{mpsc, oneshot};
-use webdriver::command::{
```
jonhoo, line 10: Why get rid of these `use`s?

src/client.rs
```diff
@@ -279,44 +279,42 @@ impl Client {
...
+                    Some(Ok(handle)) => handle,
+                    _ => return Err(error::CmdError::NotW3C(Json::Object(obj))),
                 };
 
-                let typ = match obj
-                    .remove("type")
-                    .and_then(|x| x.as_str().map(String::from))
-                {
-                    Some(typ) => typ,
+                let typ = match obj.get("type").and_then(|x| x.as_str()) {
```
jonhoo, line 296: Is that the only important field on a new window response?

src/client.rs
```diff
@@ -929,3 +927,13 @@ impl Client {
         }
     }
 }
+
+/// Response returned by [`Client::new_window()`] method.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct NewWindowResponse {
```
jonhoo, line 933: This feels like it should go in `wd`. It also feels like it should have `#[non_exhaustive]`.

src/cookies.rs
jonhoo: It also causes a _lot_ of noise in this PR.

jonhoo:
Sorry it took me so long to get back to you — life got in the way :)
The pings won't generally do much. I keep an eye on my notifications, and will get to it when I can. How recently it's been touched arguably leads to _lower_ priority since I tend to walk my notifications from oldest to newest.

Thanks (again) for all the work you're putting into this! I left more notes inline.

src/cookies.rs
tyranron: I do tend keep links as small as possible, because long ones make problem for `rustfmt`. I'll separate this change into another small PR, if the noise bothers.

src/cookies.rs
tyranron: It turns out to be a [quite subjective thing](https://github.com/rust-lang/rust-clippy/issues/2523). For me it's clearer, shorter and complies more with lints like [`clippy::redundant_closure`](https://rust-lang.github.io/rust-clippy/master/#redundant_closure), and we do use such style a lot. Though, I cannot recall Clippy explicitly complaining on it at the moment.

If you're not comfortable with it, I'm OK to revert.

src/lib.rs
tyranron: This was to make new `pub` types inside this module reachable, without necessity to polute root namespace with them (as they're quite specific).

src/wd.rs
tyranron: Just following [the spec stated in docs](https://www.w3.org/TR/webdriver/#dfn-window-handles) here:

> Each browsing context has an associated window handle which uniquely identifies it. This must be a String and must not be "current".

src/client.rs
tyranron: Across this file there was a mix: some places use long form with `webdriver::command::` prefix, other ones use short form. I've found it quite confusing. As the later was minority and the imported names weren't much, but in few places only, I've made the style consitent accross this file.

src/elements.rs
tyranron: Smae reasons as [here](https://github.com/jonhoo/fantoccini/pull/139#discussion_r767908177).

src/client.rs
tyranron: Yup. I haven't changed semantics here, the previous code using `webdriver` types was:
```rust
                let typ = match obj
                    .remove("type")
                    .and_then(|x| x.as_str().map(String::from))
                {
                    Some(typ) => typ,
                    None => return Err(error::CmdError::NotW3C(Json::Object(obj))),
                };

                Ok(webdriver::response::NewWindowResponse { handle, typ })
```

I've only typed the `typ` better [according to the spec](https://www.w3.org/TR/webdriver/#dfn-new-window):
> 7. Let type be "tab" if the newly created window shares an OS-level window with the current browsing context, or "window" otherwise.

src/client.rs
tyranron: > This feels like it should go in `wd`.

I've put it here, because it's not the type/structure explicitly declared by [the spec](https://www.w3.org/TR/webdriver), but a helper type for representing result of the concrete `Client` method. I'm used to keep such things near to the method requiring them.

Though, I'm OK with moving it to `wd` if you prefer so.

>  It also feels like it should have `#[non_exhaustive]`.

It's completely defined by [spec](https://w3c.github.io/webdriver/#dfn-new-window):
> 8. Let result be a new JSON Object initialised with:
> 
> "handle"
>     The value of handle. 
> "type"
>     The value of type. 

So it won't change unless the spec changes. And migrating to new spec is breaking change anyway.

Considering this, I think `#[non_exhaustive]` here will only create unconveniences/uncertainty for user code. Unless we want to propose here some custom extended interface, not repeat the WebDriver spec exactly.

tyranron:
@jonhoo 

> The pings won't generally do much. I keep an eye on my notifications, and will get to it when I can. How recently it's been touched arguably leads to lower priority since I tend to walk my notifications from oldest to newest.

I'm sorry, if pings were annoying. I'm just persistent at not allowing things vanishing in time 😅
Will keep in mind your workflow.

.github/dependabot.yml
tyranron: Disolved to https://github.com/jonhoo/fantoccini/pull/144

src/cookies.rs
jonhoo: Yeah, I like to keep PRs focused, otherwise reviewing gets much harder. And tracking through git history becomes more of a pain :sweat_smile: Thanks!

src/cookies.rs
jonhoo: Heh, yeah, I'm definitely more on the "drop is special in Rust and shouldn't be mixed with 'ignore' even if the outcome is the same" side of that discussion. Thanks for reverting.

src/client.rs
jonhoo: I buy both of those arguments :+1:

src/wd.rs
jonhoo: Ah, interesting. I thought a window handle could also be passed as an argument to certain commands, in which case "current" would be handled as a special case (that feels like why they'd have a carve-out for that string), but can't actually find any commands that do that at the moment, so all good then :+1:

jonhoo (approved):
LGTM — thanks again!

jonhoo:
Published as 0.18.0-beta.1 :tada:

PR #141, opened 2021-12-03
Update to hyper-rustls 0.23

complexspaces:
This PR allows downstream consumers who also depend on the `rustls` ecosystem via crates to avoid duplicate dependencies. 

I don't believe that this is a breaking change.


jonhoo:
Closing in favor of #142.

PR #142, opened 2021-12-07
Upgrade hyper-rustls to 0.23

paolobarbolini:
Properly updates to hyper-rustls 0.23, as #141 and #139 give errors when trying to compile with it.

Closes #141


jonhoo:
Unfortunately this _is_ a breaking change because the return type of [`ClientBuilder::rustls`](https://docs.rs/fantoccini/latest/fantoccini/struct.ClientBuilder.html#method.rustls) includes a type from `hyper-rustls`. However, given that I'd like to merge #139 which is a breaking change anyway, I'm okay with moving forward with this :+1:

PR #143, opened 2021-12-09
Disable the hyper http2 feature

paolobarbolini:
Since you're going to release a minor version anyway I wouldn't mind having the ability to reduce `hyper`'s dependencies.

---

Makes the `http2` feature optional and enables it by default.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,7 +30,7 @@ serde_json = "1.0.25"
 futures-core = "0.3"
 futures-util = "0.3"
 tokio = { version = "1", features = ["sync", "rt"] }
-hyper = { version = "0.14", features = ["stream", "client", "http1", "http2"] }
+hyper = { version = "0.14", features = ["stream", "client", "http1"] }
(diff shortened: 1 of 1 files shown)


jonhoo:
Good call. Should we even include http2 support at all? Is it actually supported (or more importantly, required) for any WebDriver servers?

paolobarbolini:
I'm not sure. The http2 feature was enabled in commit https://github.com/jonhoo/fantoccini/commit/715a7114d991816b84e7a9da18bb7498dfe5c2a1. I guess it was either done to explicitly get HTTP/2 enabled or to work around https://github.com/rustls/hyper-rustls/issues/141

jonhoo:
That's a very good question. @x0f5c3 can you shed some light on the use of the `http2` feature? I'd be totally happy to remove it altogether.

x0f5c3:
I think it was done to work around the rustls issue, if it was fixed I guess it could be disabled since AFAIK WebDriver protocol doesn't use HTTP2

jonhoo (approved):
Excellent, thanks!

PR #147, opened 2022-01-04
Use `try_fold` instead of `try_concat` in README

LegNeato:
`try_concat` is no longer compatible with hyper's `Body` (see https://github.com/hyperium/hyper/issues/1951). I didn't use https://docs.rs/hyper/latest/hyper/body/fn.to_bytes.html as to not require a direct hyper dependency for the snippet.


jonhoo (approved):
Excellent, thank you!

PR #151, opened 2022-01-10
Implement remaining WebDriver commands

stevepryde:
I've wired up and implemented most of the remaining `WebDriverCommand`s.

Some points of discussion:
1. A lot of methods in `fantoccini` take `&mut self` when mutable access is not required. In fact most methods should be fine with `&self`. Is this intentional?
2. In order to wrap `Element` in `thirtyfour`, I needed to make its members public. Is this going to cause any issues?
3. It would be nice to re-export some of the types used by `fantoccini`, so that calling crates can import (use) the exact versions without having to try to keep things in lockstep. I've add a couple of re-exports under `fantoccini::external` but feel free to suggest a better namespace. I'd also be fine with these being behind a feature flag in order to not pollute the namespace.
4. I haven't yet added tests (except for `add_cookie()`). I assume the desired approach is to continue with what is in tests/local.rs?

### DIFF
--- a/src/actions.rs
+++ b/src/actions.rs
@@ -2,11 +2,15 @@
 use crate::elements::Element;
 #[cfg(doc)]
 use crate::keys::Key;
+use crate::Client;
 use std::fmt::Debug;
 use std::time::Duration;
 use webdriver::actions as WDActions;
 
 /// An action not associated with an input device (i.e. pause).
+///
+/// See [17.4.1 General Actions](https://www.w3.org/TR/webdriver1/#general-actions) of the
+/// WebDriver standard.
 #[derive(Debug, Clone)]
 #[non_exhaustive]
 pub enum NullAction {
@@ -30,6 +34,9 @@ impl NullAction {
 }
 
 /// An action performed with a keyboard.
+///
+/// See [17.4.2 Keyboard Actions](https://www.w3.org/TR/webdriver1/#keyboard-actions) of the
+/// WebDriver standard.
 #[derive(Debug, Clone)]
 #[non_exhaustive]
 pub enum KeyAction {
@@ -74,8 +81,21 @@ impl KeyAction {
     }
 }
 
-/// An action performed with a pointer device. See `PointerActionType` for
-/// the availabledevice types.
+/// Left mouse button constant for use with `PointerAction`.
+pub const MOUSE_BUTTON_LEFT: u64 = 0;
+
+/// Middle mouse button constant for use with `PointerAction`.
+pub const MOUSE_BUTTON_MIDDLE: u64 = 1;
+
+/// Right mouse button constant for use with `PointerAction`.
+pub const MOUSE_BUTTON_RIGHT: u64 = 2;
+
+/// An action performed with a pointer device.
+///
+/// This can be a mouse, pen or touch device.
+///
+/// See [17.4.3 Pointer Actions](https://www.w3.org/TR/webdriver1/#pointer-actions) of the
+/// WebDriver standard.
 #[derive(Debug, Clone)]
 #[non_exhaustive]
 pub enum PointerAction {
@@ -101,11 +121,12 @@ pub enum PointerAction {
         /// - Right = 2
         button: u64,
     },
-    /// Pointer move action. Duration is specified in milliseconds (can be 0).
+    /// Pointer move action.
+    ///
     /// The x and y offsets are relative to the origin.
     Move {
-        /// The move duration, given in milliseconds.
-        duration: u64,
+        /// The move duration.
+        duration: Option<Duration>,
         /// The origin that the `x` and `y` coordinates are relative to.
         origin: PointerOrigin,
         /// `x` offset, in pixels.
@@ -134,31 +155,9 @@ impl PointerOrigin {
         match self {
             PointerOrigin::Viewport => WDActions::PointerOrigin::Viewport,
             PointerOrigin::Pointer => WDActions::PointerOrigin::Pointer,
-            PointerOrigin::WebElement(e) => {
-                WDActions::PointerOrigin::Element(webdriver::common::WebElement(e.element_id().0))
-            }
-        }
-    }
-}
-
-/// The type of pointer.
-#[derive(Debug, Clone)]
-#[non_exhaustive]
-pub enum PointerActionType {
-    /// A mouse pointer device.
-    Mouse,
-    /// A pen device.
-    Pen,
-    /// A touch device.
-    Touch,
-}
-
-impl PointerActionType {
-    fn into_wd_pointertype(self) -> WDActions::PointerType {
-        match self {
-            PointerActionType::Mouse => WDActions::PointerType::Mouse,
-            PointerActionType::Pen => WDActions::PointerType::Pen,
-            PointerActionType::Touch => WDActions::PointerType::Touch,
+            PointerOrigin::WebElement(e) => WDActions::PointerOrigin::Element(
+                webdriver::common::WebElement(e.element_id().into()),
+            ),
         }
     }
 }
@@ -184,7 +183,7 @@ impl PointerAction {
                 y,
             } => WDActions::PointerActionItem::Pointer(WDActions::PointerAction::Move(
                 WDActions::PointerMoveAction {
-                    duration: Some(duration),
+                    duration: duration.map(|x| x.as_millis() as u64),
                     origin: origin.into_wd_pointerorigin(),
                     x: Some(x),
                     y: Some(y),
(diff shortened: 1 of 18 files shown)


jonhoo:
Thanks for getting started on this!

> 1. A lot of methods in `fantoccini` take `&mut self` when mutable access is not required. In fact most methods should be fine with `&self`. Is this intentional?

The intention was to model what the browser context through the ownership system, so any method that changes the browser context (go to URL, set cookie, exec JavaScript, etc.) take `&mut self`, whereas anything that's read-only takes `&self`. I don't know how much that buys us in practice though. If you feel like we'd be better off sticking with one or the other, I'm okay with that.

> 2. In order to wrap `Element` in `thirtyfour`, I needed to make its members public. Is this going to cause any issues?

Ah, yes, so, this is actually somewhat complicated. We _explicitly_ removed all types from the `webdriver` crate from public APIs in `fantoccini` (https://github.com/jonhoo/fantoccini/pull/139) so that we don't have to issue a new major version of `fantoccini` any time the `webdriver` crate issues a new breaking release (which they tend to do rather frequently). There's more discussion in the linked PR, but essentially we should be able to cut a 1.0 of fantoccini and then continue to upgrade `webdriver` under the hood without having to release a (X+1).0 each time we do. _Any_ public exposure of a type from `webdriver` means a bump to `webdriver` would be backwards-incompatible.

My suggestion is therefore to instead add accessors for specifically the properties you need, rather than making `webdriver`-typed fields public.

> 3. It would be nice to re-export some of the types used by `fantoccini`, so that calling crates can import (use) the exact versions without having to try to keep things in lockstep. I've add a couple of re-exports under `fantoccini::external` but feel free to suggest a better namespace. I'd also be fine with these being behind a feature flag in order to not pollute the namespace.

As above, the idea is to make `webdriver` not be a public dependency of `fantoccini` so that clients _don't_ have to upgrade in lock-step. I see you re-exported `url` which I think I'm more okay with since it tends to be more stable release-wise, but I'm not sure `url` on its own buys us that much? It should be fairly rare that someone prefers `fantoccini::externals::url::Url` over `url::Url` with a `Cargo.toml` dependency.

> 4. I haven't yet added tests (except for `add_cookie()`). I assume the desired approach is to continue with what is in tests/local.rs?

Yes, in general tests in the style of `tests/local.rs` are preferable since they don't rely on external sources that can change outside of our control, and also don't require network connectivity. Ideally I'd like to move all the tests to be local. Now, we may want to split `tests/local.rs` into multiple files so it doesn't end up being huge, but that's a detail.

stevepryde:
> > 1. A lot of methods in `fantoccini` take `&mut self` when mutable access is not required. In fact most methods should be fine with `&self`. Is this intentional?
> 
> The intention was to model what the browser context through the ownership system, so any method that changes the browser context (go to URL, set cookie, exec JavaScript, etc.) take `&mut self`, whereas anything that's read-only takes `&self`. I don't know how much that buys us in practice though. If you feel like we'd be better off sticking with one or the other, I'm okay with that.

I wondered if that was the case. In my opinion that probably only makes sense in a few cases such as whole-page navigation. Web applications these days are highly dynamic so even switching to a frame would only invalidate element ids temporarily until you switch back to the original window context. Things like the element click taking ownership also seems quite heavy-handed to me. 90% of the clicks on a page would not invalidate most of the page. It's more common nowadays for only individual components to re-render, and it's probably not feasible for `fantoccini` to know/decide when that happens, and so my opinion is that we shouldn't try.

Calling code still needs to handle element staleness regardless (which `fantoccini` bundles with `NoSuchElement`, and that's probably fine in practice). So my preference would be to stick with immutable references unless there's a clear reason not to (such as `goto()`?).

That said, this change doesn't need to be in this PR, so I'm happy to follow the existing mutability scheme and we can look at changing to immutable references later. Thoughts?

> > 2. In order to wrap `Element` in `thirtyfour`, I needed to make its members public. Is this going to cause any issues?
> 
> Ah, yes, so, this is actually somewhat complicated. We _explicitly_ removed all types from the `webdriver` crate from public APIs in `fantoccini` (#139) so that we don't have to issue a new major version of `fantoccini` any time the `webdriver` crate issues a new breaking release (which they tend to do rather frequently). There's more discussion in the linked PR, but essentially we should be able to cut a 1.0 of fantoccini and then continue to upgrade `webdriver` under the hood without having to release a (X+1).0 each time we do. _Any_ public exposure of a type from `webdriver` means a bump to `webdriver` would be backwards-incompatible.
> 
> My suggestion is therefore to instead add accessors for specifically the properties you need, rather than making `webdriver`-typed fields public.

Good point! I'll do that instead :)
I will likely need a constructor for `Element` as well.
And if this intent is not documented anywhere then I might look at adding something somewhere for future travelers.
 
> > 3. It would be nice to re-export some of the types used by `fantoccini`, so that calling crates can import (use) the exact versions without having to try to keep things in lockstep. I've add a couple of re-exports under `fantoccini::external` but feel free to suggest a better namespace. I'd also be fine with these being behind a feature flag in order to not pollute the namespace.
> 
> As above, the idea is to make `webdriver` not be a public dependency of `fantoccini` so that clients _don't_ have to upgrade in lock-step. I see you re-exported `url` which I think I'm more okay with since it tends to be more stable release-wise, but I'm not sure `url` on its own buys us that much? It should be fairly rare that someone prefers `fantoccini::externals::url::Url` over `url::Url` with a `Cargo.toml` dependency.

Agreed. I think it's fine to remove both then.
 
> > 4. I haven't yet added tests (except for `add_cookie()`). I assume the desired approach is to continue with what is in tests/local.rs?
> 
> Yes, in general tests in the style of `tests/local.rs` are preferable since they don't rely on external sources that can change outside of our control, and also don't require network connectivity. Ideally I'd like to move all the tests to be local. Now, we may want to split `tests/local.rs` into multiple files so it doesn't end up being huge, but that's a detail.

Cool. Sounds like a plan :)

stevepryde:
So in order to not expose `webdriver` types I assume that means creating new types that can convert to them behind the scenes.

I've done that for the timeouts (`get_timeouts()` and `set_timeouts()`) but the action chains are pretty complex and difficult to coerce into an interface that is both easy to use as well as retaining all the flexibility of the underlying spec.

I could remove the `perform_actions()` and `release_actions()` and look at those later?
Or if you have some idea of how to implement it then I'd be open to suggestions. I have some ideas for a fairly high-level interface but as always flexibility and ease-of-use tend to be opposing constraints.

stevepryde:
Never mind - I've figured out a good solution for actions. I will update the PR for feedback soon.

stevepryde:
I've updated the code including various fixes after testing with the `thirtyfour` test suite. All of those tests are now passing there so I have some level of confidence that things are working, but `fantoccini` still needs its own tests.

The `ActionChain` stuff is fairly close to what the actual webdriver supports. I believe any combination of actions is possible, and hopefully the interface also makes sense. I had a bunch of tests already for this and they are all passing. The unit tests have been moved over, but the integration tests still need to be ported.

AFAIK the only thing left to do is add tests for everything, which I plan to do.
If you could review the code that's there sometime and let me know if any changes are required that would be great. No rush, just when you get time. I'll chip away at the tests in the meantime.

src/client.rs
```diff
@@ -200,6 +250,19 @@ impl Client {
...
         Ok(())
     }
+
+    /// Get the current page title.
+    ///
+    /// See [9.6 Get Title](https://www.w3.org/TR/webdriver1/#dfn-get-title) of the WebDriver standard.
+    #[cfg_attr(docsrs, doc(alias = "Get Title"))]
+    pub async fn title(&mut self) -> Result<String, error::CmdError> {
+        let title = self.issue(WebDriverCommand::GetTitle).await?;
+        if let Some(title) = title.as_str() {
```
jonhoo, line 260: Let's use `let Some(Value::String(s)) = title { return s }` here instead to avoid unnecessarily allocating a new string :+1:

src/client.rs
```diff
@@ -600,6 +690,91 @@ impl Client {
...
+    ///     .pause(Duration::from_secs(2))
+    ///     .send_keys("fantoccini")
+    ///     .build();
+    /// client.perform_actions(chain).await?;
+    /// ```
+    ///
+    /// See [17.5 Perform Actions](https://www.w3.org/TR/webdriver1/#perform-actions) of the
+    /// WebDriver standard.
+    #[cfg_attr(docsrs, doc(alias = "Perform Actions"))]
+    pub async fn perform_actions(&mut self, actions: ActionChain) -> Result<(), error::CmdError> {
```
jonhoo, line 711: What do you think of adapting this slightly to have
```rust
impl Client {
  fn action_chain(&mut self) -> ActionChain;
}
impl ActionChain {
  fn perform(self) -> Result<(), error::CmdError>;
}
```
(where `ActionChain` here is the same as the builder type you currently have). I _think_ that'd make for a nicer interface?

src/client.rs
```diff
@@ -600,6 +690,91 @@ impl Client {
...
+    }
+
+    /// Get the text of the active alert, if there is one.
+    ///
+    /// See [18.3 Get Alert Text](https://www.w3.org/TR/webdriver1/#get-alert-text) of the
+    /// WebDriver standard.
+    #[cfg_attr(docsrs, doc(alias = "Get Alert Text"))]
+    pub async fn get_alert_text(&mut self) -> Result<String, error::CmdError> {
+        let res = self.issue(WebDriverCommand::GetAlertText).await?;
+        if let Some(v) = res.as_str() {
```
jonhoo, line 757: Same here re re-using `String`

src/cookies.rs
```diff
@@ -1,25 +1,109 @@
...
 
 /// Type alias for a [cookie::Cookie]
 pub type Cookie<'a> = cookie::Cookie<'a>;
 
+/// Wrapper for serializing AddCookieParameters.
+#[derive(Debug, Serialize)]
+pub struct AddCookieParametersWrapper<'a> {
+    /// The cookie to serialize.
+    #[serde(with = "AddCookieParameters")]
+    pub cookie: &'a AddCookieParameters,
```
jonhoo, line 18: This would make `webdriver` a public dependency again, so we can't have this.

src/cookies.rs
```diff
@@ -1,25 +1,109 @@
...
+    /// or frames into a third party site), but are sent when a user is navigating to the
+    /// origin site (i.e., when following a link).
+    Lax,
+    /// Cookies will be sent in all contexts, i.e. in responses to both first-party and
+    /// cross-origin requests. If SameSite=None is set, the cookie Secure attribute must
+    /// also be set (or the cookie will be blocked).
+    None,
+}
+
+impl From<String> for SameSite {
```
jonhoo, line 40: Should this be `FromStr` instead? Or `From<&str>`? Seems odd to constrain it to `String`.

src/cookies.rs
```diff
@@ -1,25 +1,109 @@
...
+    fn from(s: String) -> Self {
+        match s.to_lowercase().as_ref() {
+            "strict" => SameSite::Strict,
+            "lax" => SameSite::Lax,
+            _ => SameSite::None,
```
jonhoo, line 45: This clause makes me think this should be `TryFrom` (or `FromStr`), and not just `From`.

src/cookies.rs
```diff
@@ -1,25 +1,109 @@
...
+        }
+    }
+}
+
+impl From<SameSite> for String {
```
jonhoo, line 50: I think this is better expressed as `impl fmt::Display for SameSite` instead.

src/cookies.rs
```diff
@@ -1,25 +1,109 @@
...
+    #[serde(with = "AddCookieParameters")]
+    pub cookie: &'a AddCookieParameters,
+}
+
+/// The SameSite parameter for cookies.
+///
+/// See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite
+/// for more details.
+#[derive(Debug, Clone, Copy, Serialize, Deserialize)]
+pub(crate) enum SameSite {
```
jonhoo, line 26: Can we not re-use `cookie::SameSite` here? Why do we need our own type when it's only `pub(crate)`?

src/elements.rs
```diff
@@ -15,24 +15,40 @@ use webdriver::error::WebDriverError;
...
-#[derive(Clone, Debug)]
-pub struct Form {
-    pub(crate) client: Client,
-    pub(crate) form: webdriver::common::WebElement,
-}
-
 impl Element {
+    /// Construct an `Element` with the specified element id.
+    /// The element id is the id given by the webdriver.
+    pub fn from_element_id(client: Client, element_id: String) -> Self {
```
jonhoo, line 29: I wonder if we should provide a newtype for `ElementId` given how common it is in the API?

src/elements.rs
```diff
@@ -107,6 +123,45 @@ impl Element {
...
 impl Element {
+    /// Return true if the element is currently selected.
+    ///
+    /// See [13.1 Is Element Selected](https://www.w3.org/TR/webdriver1/#is-element-selected)
+    /// of the WebDriver standard.
+    pub async fn is_selected(&mut self) -> Result<bool, error::CmdError> {
+        let cmd = WebDriverCommand::IsSelected(self.element.clone());
+        match self.client.issue(cmd).await? {
+            Json::Bool(v) => Ok(v),
+            Json::Null => Ok(false),
```
jonhoo, line 134: Is this valid according to the spec? If not, we should leave a comment about why it's needed. And same for the other two instances below.

src/elements.rs
```diff
@@ -139,12 +194,33 @@ impl Element {
         let cmd = WebDriverCommand::GetElementProperty(self.element.clone(), prop.to_string());
         match self.client.issue(cmd).await? {
             Json::String(v) => Ok(Some(v)),
+            Json::Bool(b) => {
+                // NOTE: boolean properties such as "checked" seem to be returned as boolean
+                //       values rather than strings.
```
jonhoo, line 199: That's super interesting. I guess it's because you can have something like `<option selected>`. Though I thought that was shorthand for `<option selected="selected">`. Not quite sure what the right thing to do here is in terms of coercing to a string type. Whatever we do ("true" and "false" are probably okay), we need to document that in the public API, not just in an internal comment.

src/elements.rs
```diff
@@ -139,12 +194,33 @@ impl Element {
...
+            }
             Json::Null => Ok(None),
             v => Err(error::CmdError::NotW3C(v)),
         }
     }
 
-    /// Retrieve the text contents of this elment.
+    /// Look up a CSS property for this element by name.
+    ///
+    /// `Ok(String::new())` is returned if the the given CSS property is not found.
```
jonhoo, line 209: Why not use `Option` here same as for `prop` above?

src/elements.rs
```diff
@@ -157,6 +233,57 @@ impl Element {
...
+            v => Err(error::CmdError::NotW3C(v)),
+        }
+    }
+
+    /// Gets the x, y, width, and height properties of the current element.
+    ///
+    /// See [13.7 Get Element Rect](https://www.w3.org/TR/webdriver1/#dfn-get-element-rect) of the
+    /// WebDriver standard.
+    #[cfg_attr(docsrs, doc(alias = "Get Element Rect"))]
+    pub async fn get_element_rect(&mut self) -> Result<(f64, f64, u64, u64), error::CmdError> {
```
jonhoo, line 254: I don't think the spec says anything about `x` and `y` being fractional while `width` and `height` are whole numbers. They all appear to be "CSS Pixel" values, and I _think_ those can always be fractional, in which case all of these should be `f64`.

jonhoo: Also, since this is already a method _on_ `Element`, I think it should be called `get_rect`, or better yet just `rect`. Or why not properly: `rectangle`? The point f the `doc(alias` is specifically so that we can name the methods something else if we disagree with the standard, and still have it be searchable by the standard name.

src/keys.rs
```diff
@@ -0,0 +1,194 @@
+//! Key codes for use with Actions.
+
+use std::ops::Add;
+
+/// Key codes for use with Actions.
+#[derive(Debug)]
+pub enum Keys {
```
jonhoo, line 7: I think this should just be `Key`, since it can only represent a single key.

src/keys.rs
```diff
@@ -0,0 +1,194 @@
...
+            Keys::F10 => '\u{e03a}',
+            Keys::F11 => '\u{e03b}',
+            Keys::F12 => '\u{e03c}',
+            Keys::Meta => '\u{e03d}',
+            Keys::Command => '\u{e03d}',
+        }
+    }
+}
+
+impl<S> Add<S> for Keys
```
jonhoo, line 185: This would end up very inefficient in practice. How about instead:
```rust
impl Add<Key> for String {
```
and similarly for `AddAssign`?

src/wd.rs
```diff
@@ -1,15 +1,19 @@
...
 use std::fmt;
+use std::time::Duration;
+use webdriver::command::TimeoutsParameters;
 
 /// A [handle][1] to a browser window.
 ///
 /// Should be obtained it via [`Client::window()`] method (or similar).
 ///
 /// [1]: https://www.w3.org/TR/webdriver/#dfn-window-handles
+/// [`Client::window()`]: crate::Client::window
```
jonhoo, line 16: You can avoid the need for all of these, I think, by adding this to the `use` list:
```rust
#[cfg(doc)]
use crate::Client;
```

src/wd.rs
```diff
@@ -156,3 +160,118 @@ impl<'a> Locator<'a> {
...
+        self.implicit.map(Duration::from_millis)
+    }
+
+    /// Set the implicit wait timeout.
+    pub fn set_implicit(&mut self, timeout: Option<Duration>) {
+        self.implicit = timeout.map(|x| x.as_millis() as u64);
+    }
+}
+
+impl From<TimeoutsParameters> for TimeoutConfiguration {
```
jonhoo, line 259: Unfortunately even this leaks the `webdriver` crate as a public part of `fantoccini`'s API since all trait impls are public. You'll need to have a private inherent conversion method instead. We'll also have to do that for the other adapter types — none of the ones that are `pub` can have trait impls related to `webdriver` types. It's awkward, but unfortunately until we get something like [hidden trait implementations](https://github.com/rust-lang/rfcs/pull/2529) that's what's needed. And until we get [private dependencies](https://github.com/rust-lang/rust/issues/44663), it's hard to catch these too :sweat_smile:

tests/remote.rs
```diff
@@ -101,7 +102,11 @@ async fn raw_inner(mut c: Client) -> Result<(), error::CmdError> {
     let src = img.attr("src").await?.expect("image should have a src");
 
     // now build a raw HTTP client request (which also has all current cookies)
-    let raw = img.client().raw_client_for(Method::GET, &src).await?;
+    let raw = img
+        .client()
+        .clone()
```
jonhoo, line 107: I'm not sure I follow why this now requires a `clone`?

src/actions.rs
```diff
@@ -0,0 +1,886 @@
...
+#[derive(Debug, Clone)]
+pub enum NullAction {
+    /// Pause for the specified duration.
+    Pause {
+        /// The pause duration, given in milliseconds.
+        duration: u64,
+    },
+}
+
+impl From<NullAction> for WDActions::NullActionItem {
```
jonhoo, line 17: Unfortunately all of these `From` impls must be replaced with private inherent methods instead. See my comment elsewhere for why.

src/actions.rs
```diff
@@ -0,0 +1,886 @@
+//! Actions functionality for WebDriver.
+use crate::elements::Element;
+use std::fmt::Debug;
+use std::time::Duration;
+use webdriver::actions as WDActions;
+
+/// An action not associated with an input device (i.e. pause).
+#[derive(Debug, Clone)]
+pub enum NullAction {
```
jonhoo, line 9: I suspect we'll want most of these to be `#[non_exhaustive]`. I don't think it costs us much, since these won't generally be matched on externally anyway, and it'll allow us to add new actions without doing a breaking release.

src/actions.rs
```diff
@@ -0,0 +1,886 @@
...
+}
+
+/// An action performed with a keyboard.
+#[derive(Debug, Clone)]
+pub enum KeyAction {
+    /// Pause action.
+    /// Useful for adding pauses between other key actions.
+    Pause {
+        /// The pause duration, given in milliseconds.
+        duration: u64,
```
jonhoo, line 36: Can this be `Duration`?

src/actions.rs
```diff
@@ -0,0 +1,886 @@
...
+    },
+    /// Key up action.
+    Up {
+        /// The key code, e.g. `'a'`. See the [`Keys`] module for special key codes.
+        ///
+        /// [`Keys`]: crate::keys::Keys
+        value: char,
```
jonhoo, line 43: Should we have
```rust
enum KeyCode {
  Special(Key),
  Type(char)
}
```
or something like it, and then use that enum type here?

src/actions.rs
```diff
@@ -0,0 +1,886 @@
...
+pub enum PointerAction {
+    /// Pause action.
+    /// Useful for adding pauses between other key actions.
+    Pause {
+        /// The pause duration, given in milliseconds.
+        duration: u64,
+    },
+    /// Pointer button down.
+    Down {
+        /// The mouse button index. Button values are as follows:
```
jonhoo, line 89: Can this be an `enum`?

src/actions.rs
```diff
@@ -0,0 +1,886 @@
...
+        /// - Left = 0
+        /// - Middle = 1
+        /// - Right = 2
+        button: u64,
+    },
+    /// Pointer move action. Duration is specified in milliseconds (can be 0).
+    /// The x and y offsets are relative to the origin.
+    Move {
+        /// The move duration, given in milliseconds.
+        duration: u64,
```
jonhoo, line 107: `Duration`?

src/actions.rs
```diff
@@ -0,0 +1,886 @@
...
+        /// The origin that the `x` and `y` coordinates are relative to.
+        origin: PointerOrigin,
+        /// `x` offset, in pixels.
+        x: i64,
```
jonhoo, line 111: Do we know that pixel values must be integers (and not floats)?

src/actions.rs
```diff
@@ -0,0 +1,886 @@
...
+            }
+        }
+    }
+}
+
+/// A channel containing `Null` actions.
+#[derive(Debug, Clone)]
+pub struct NullActionChannel {
+    /// An identifier to distinguish this channel from others. Choose a meaningful string.
+    /// May be useful for debugging.
```
jonhoo, line 201:
```suggestion
    /// An identifier to distinguish this channel from others.
    ///
    /// Choose a meaningful string as it may be useful for debugging.
```
Is going to render more nicely.

src/actions.rs
```diff
@@ -0,0 +1,886 @@
...
+            PointerAction::Cancel => {
+                WDActions::PointerActionItem::Pointer(WDActions::PointerAction::Cancel)
+            }
+        }
+    }
+}
+
+/// A channel containing `Null` actions.
+#[derive(Debug, Clone)]
+pub struct NullActionChannel {
```
jonhoo, line 199: Are these called "channels" in the spec too? Otherwise, might I suggest `NullActions`?

src/actions.rs
```diff
@@ -0,0 +1,886 @@
...
+    /// Create a new NullActionChannel.
+    pub fn new(id: &str) -> Self {
+        Self {
+            id: id.to_string(),
+            actions: Vec::new(),
+        }
+    }
+
+    /// Add a pause action to this channel.
+    pub fn add_pause(&mut self, duration: Duration) {
```
jonhoo, line 217: Might it make sense to call this (these) just `pause`? I think `add_` is implied.

src/actions.rs
```diff
@@ -0,0 +1,886 @@
...
+        self.add_action(NullAction::Pause {
+            duration: duration.as_millis() as u64,
+        });
+    }
+
+    /// Add the specified action to this channel.
+    pub fn add_action(&mut self, action: NullAction) {
```
jonhoo, line 224: How about using the name `then` for this (these)?

jonhoo: Also, let's return `&mut Self` from each of these so that they're easier to construct by chaining.

src/actions.rs
```diff
@@ -0,0 +1,886 @@
...
+/// will be executed simultaneously, then all of the second actions, and so on.
+/// The second column will not be executed until all actions in the first column have been
+/// completed, including any pauses.
+///
+/// Thus, the duration of each column will be equal to the longest duration of any individual
+/// action in that column.
+#[derive(Debug, Clone)]
+pub enum ActionChannel {
+    /// A channel containing null actions.
+    Null(NullActionChannel),
```
jonhoo, line 308: Is the "Null" naming dictated by the spec? Feels like a very non-descript word for what these actions really represent.

src/actions.rs
```diff
@@ -0,0 +1,886 @@
...
+    }
+
+    /// Add the specified action to this channel.
+    pub fn add_action(&mut self, action: PointerAction) {
+        self.actions.push(action);
+    }
+}
+
+/// An ActionChannel is a sequence of actions of a specific type.
+/// When combined with other channels, you can think of them like a grid, with actions on the
```
jonhoo, line 297: If you have the time, and ASCII diagram would be _amazing_ here!

Also, I think convention is to leave a blank line between the summary of a method/type (the first sentence) and what follows.

src/actions.rs
```diff
@@ -0,0 +1,886 @@
...
+            sequences.push(channel.into_sequence());
+        }
+
+        webdriver::command::ActionsParameters { actions: sequences }
+    }
+}
+
+/// Builder for an ActionChain. Use `ActionChain::builder()` to create one.
+#[derive(Debug)]
+pub struct ActionChainBuilder {
```
jonhoo, line 394: This feels like something that should live in `thirtyfour`, and not `fantoccini`.

jonhoo (changes requested):
Phew, okay, had some time today to dig through this. A lot of good stuff in here, thanks for taking the time!

src/client.rs
stevepryde: I agree, but currently the builder isn't really capable of covering everything that the `Actions` spec supports. It doesn't handle touch/pen actions, multiple of the same type, or chords. However I can probably just add builder methods to allow the user to pass in an entire channel, which should fill the gap.

This does conflict with your other suggestion to move `ActionChainBuilder` into `thirtyfour`, which I think might be the better option but I don't want to leave the `fantoccini` side feeling too low-level to be useful.

src/cookies.rs
stevepryde: Ah, yep I missed the fact that traits are public.

src/cookies.rs
stevepryde: `cookie::SameSite` doesn't implement `Serialize` or `Deserialize`. But I can do this another way.

src/cookies.rs
stevepryde: Removed `SameSite`

src/cookies.rs
stevepryde: Removed `SameSite`

src/cookies.rs
stevepryde: Removed `SameSite`

src/elements.rs
stevepryde: Good idea!

src/elements.rs
stevepryde: Yes I believe it is valid. Which part were you unsure of? The inclusion of `IsSelected` ? or the boolean value?

Here's the spec: https://www.w3.org/TR/webdriver1/#is-element-selected
(which is already mentioned in the docs here)
```
Let selected be the value corresponding to the first matching statement:

element is an input element with a type attribute in the Checkbox- or Radio Button state

    The result of element’s checkedness. 
element is an option element

    The result of element’s selectedness. 
Otherwise
    False. 
```
Those are all boolean. And, this is tested in the `thirtyfour` doctest.

src/elements.rs
stevepryde: Pretty sure this has caught me out twice. In this case I found it due to the `thirtyfour` doctest failing.
I've moved the comment into the public API docs.

src/elements.rs
stevepryde: It's tricky. The spec is different for this one, and it probably never actually returns `null`. 

```
Let computed value be the result of the first matching condition:

current browsing context’s document type is not "xml"
    computed value of parameter property name from element’s style declarations. property name is obtained from url variables. 
Otherwise
    "" (empty string) 
```

I noticed this when testing `thirtyfour`. If you query a CSS property that doesn't exist, you don't get null/None. You get an empty string.

I don't really like the idea of changing it to `Option<String>` if it never actually returns `None` in practice. However the alternatives are to leave it like this, or otherwise remove the `Json::Null` variant and let that fall through to `NotW3C`.

What's your preference?

src/elements.rs
stevepryde: As for the types, I noticed by experimentation that `x.as_u64()` returned `None`, because it was a float. The width and height seemed to always be integers.

Happy to make them all f64.

src/keys.rs
stevepryde: Ok, but wouldn't you want the ability to specify `Key::Control + "a"` for example?

src/wd.rs
stevepryde: Agreed. Done.

tests/remote.rs
stevepryde: It doesn't. It's leftover from a previous change.

src/actions.rs
stevepryde: What would this buy us? `Key` converts into `char` anyway

src/actions.rs
stevepryde: Hmm, not really. We could provide an enum for it, but AFAIK there's no actual limit to the number of buttons a mouse could have. Also I'm not sure the same enum would even make sense for other pointer types.

An enum with just `Left`, `Middle`, `Right` might be useful, but I think it should still store the button as `u64`.

src/actions.rs
stevepryde: The `webdriver` type `PointerMoveAction` has them all as i64, so that's all we've got.

src/actions.rs
stevepryde: It's used in both the python selenium library as well as the `webdriver` types.
They also use `GeneralAction` for the pause action so that might be nicer?

src/actions.rs
stevepryde: I'll give it a go. The "channel" idea is just something I came up with to try to visualize it better. If you think a different term would fit better I'm open to it.

stevepryde: Actually the spec has `input sources` so I could rename `Channel` to `InputSource` everywhere if that is preferred.

src/actions.rs
stevepryde: I'm open to that. See the above discussion in relation to `Client::perform_actions()` though.

src/client.rs
jonhoo: Yeah, I think I changed my mind on this — I think we should leave the builder out of `fantoccini`. As we talked about earlier, `fantoccini` _should_ be a low-level interface, and an action builder feels antithetical to that. It also might mean that we can more easily add support for the bits of actions that aren't yet supported, since we don't _also_ have to come up with a builder interface for it.

src/actions.rs
jonhoo: The diagram looks great!

Ah, interesting. In that case, how about:
```rust
// Needed just so we don't expose a `webdriver` type in the public API
pub struct ActionSequence(WDActions::ActionSequence);

trait InputSource: Into<ActionSequence> {
    type Action;
    fn pause(&mut self, Duration) -> &mut Self;
    fn then(&mut self, Self::Action) -> &mut Self;
}

impl InputSource for NullActions { /* .. */ }
impl InputSource for KeyActions { /* .. */ }
impl InputSource for TouchActions {
    type Action = PointerAction;
    // ...
}
impl InputSource for MouseActions {
    type Action = PointerAction;
    // ...
}
impl InputSource for PenActions {
    type Action = PointerAction;
    // ...
}

pub struct Actions {
    sequences: Vec<ActionSequence>
}

impl Actions {
    fn and(&mut self, sequence: impl Into<ActionSequence>) -> &mut Self {
        self.sequences.push(sequence.into());
    }
}
```

src/actions.rs
jonhoo: I think it'd be helpful to link to the relevant spec section for each input source type.

src/actions.rs
jonhoo: Ah, awkward. But fair enough.

src/actions.rs
jonhoo: How about including
```rust
pub type MouseButtonLeft = 0;
pub type MouseButtonMiddle = 1;
pub type MouseButtonRight = 2;
```
?

src/actions.rs
jonhoo: It would avoid the need to call `Key::into()`, but yeah, point taken :+1:

src/keys.rs
jonhoo: Good point. Maybe include all of those then? So that:
```rust
Key::Control + Key::Meta + "a"
```
will also work.

src/elements.rs
jonhoo: Yeah, let's have them all be floats to cover our bases here given that the spec claims they all have the same unit :+1:

src/elements.rs
```diff
@@ -8,6 +8,10 @@ use webdriver::command::WebDriverCommand;
 use webdriver::common::FrameId;
 use webdriver::error::WebDriverError;
 
+/// An element id.
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct ElementId(pub String);
```
jonhoo, line 13: We should also implement `Display`, `AsRef<str>`, and `Deref<str>` for this to make it a little more convenient to use. And `#[derive(Hash, Ord, PartialOrd)]`.

src/elements.rs
```diff
@@ -8,6 +8,10 @@ use webdriver::command::WebDriverCommand;
 use webdriver::common::FrameId;
 use webdriver::error::WebDriverError;
 
+/// An element id.
```
jonhoo, line 11: Is there a part of the spec we can link to for the definition of an element ID?

Also, are there other places in the API where this type may be useful?

src/cookies.rs
```diff
@@ -132,11 +79,10 @@ impl From<WebDriverCookie> for Cookie<'static> {
         }
 
         if let Some(same_site) = webdriver_cookie.same_site {
-            use cookie::SameSite as CookieSameSite;
-            cookie.set_same_site(match same_site {
-                SameSite::Strict => CookieSameSite::Strict,
-                SameSite::Lax => CookieSameSite::Lax,
-                SameSite::None => CookieSameSite::None,
+            cookie.set_same_site(match same_site.as_ref() {
```
jonhoo, line 82: You could probably make this a little nicer with [`eq_ignore_ascii_case`](https://doc.rust-lang.org/std/primitive.str.html#method.eq_ignore_ascii_case)

src/cookies.rs
```diff
@@ -132,11 +79,10 @@ impl From<WebDriverCookie> for Cookie<'static> {
...
+                "STRICT" | "Strict" | "strict" => SameSite::Strict,
+                "NONE" | "None" | "none" => SameSite::None,
+                _ => SameSite::Lax,
```
jonhoo, line 85: I think we should make this specifically check for `"lax"`, and panic if it's something we don't recognize. Or, alternatively, turn this into a `impl TryFrom` instead.

src/cookies.rs
```diff
@@ -12,51 +13,12 @@ pub type Cookie<'a> = cookie::Cookie<'a>;
 
 /// Wrapper for serializing AddCookieParameters.
 #[derive(Debug, Serialize)]
-pub struct AddCookieParametersWrapper<'a> {
+pub(crate) struct AddCookieParametersWrapper<'a> {
     /// The cookie to serialize.
     #[serde(with = "AddCookieParameters")]
     pub cookie: &'a AddCookieParameters,
```
jonhoo, line 19: This should probably be `pub(crate)`

src/elements.rs
jonhoo: And probably also `From<String>`

src/elements.rs
jonhoo: Ah, sorry, no, I mean the `Json::Null`. Is it possible for this API to return a `null` at all? If it's not, we should return that as `NotW3C`.

src/elements.rs
jonhoo: Interesting! In that case, yeah, let's leave the return type as-is, but change `Null` (which it sounds like can't be returned) to be `NotW3C` instead.

src/client.rs
stevepryde: So would you leave `Client::perform_actions()` as is then?

src/elements.rs
stevepryde: Should this be called `css_prop` instead?

src/actions.rs
stevepryde: I'm ok with this. I'd also be ok with an enum with `#[repr(u64)]` but it's slightly nicer to not have to type `as u64` every time. Let's go with the above for now.

stevepryde: Do you mean `pub const` ? This is fine. An enum with `#[repr(u64)]` would also be fine, but slightly annoying to require `as u64` each time.

As a `const` should it be `MOUSE_BUTTON_LEFT` etc?

src/actions.rs
stevepryde: Very nice! Thanks!

stevepryde:
Regarding `PointerActions`, see the following:
```rust
    /// Pointer move action.
    ///
    /// The x and y offsets are relative to the origin.
    Move {
        /// The move duration.
        duration: Option<Duration>,
        /// The origin that the `x` and `y` coordinates are relative to.
        origin: PointerOrigin,
        /// `x` offset, in pixels.
        x: i64,
        /// `y` offset, in pixels.
        y: i64,
    },
```

Is it worth factoring out `PointerOrigin` and doing something like the following?
```rust
MoveBy { duration, x, y }
MoveTo { duration, x, y }
MoveToElement { element, duration, x, y }
```

src/actions.rs
stevepryde: Removed.

src/elements.rs
stevepryde: Kind of. There's [11. Elements](https://www.w3.org/TR/webdriver1/#elements) which lists this as the second item:
```
The web element identifier is the string constant "element-6066-11e4-a52e-4f735466cecf". 
```

stevepryde: As for where else it might be useful, I'm not sure. It's definitely useful for constructing an element or any kind of (de)serialization involving elements.

c43721:
Can I chime into this and say what a great job both of you are doing here. I appreciate the time and effort being put into this, and I cannot wait to see what new things are going to be built on top of this. 

Again, a huge respect to this massive endeavour!

src/actions.rs
```diff
@@ -0,0 +1,497 @@
...
+        self.then(PointerAction::Pause { duration })
+    }
+
+    fn then(mut self, action: Self::Action) -> Self {
+        self.actions.push(action);
+        self
+    }
+}
+
+/// A list of action sequences to be performed via `Client::perform_actions()`
```
jonhoo, line 452: `perform_actions` has been renamed to just `perform`. If you make this
```suggestion
/// A list of action sequences to be performed via [`Client::perform()`]
```
It should be auto-linked by rustdoc, and it'll also tell us if the name ever changes in the future.

src/client.rs
```diff
@@ -600,6 +690,114 @@ impl Client {
...
+    ///         y: 0,
+    ///     })
+    ///     .then(PointerAction::Up {
+    ///         button: MOUSE_BUTTON_LEFT,
+    ///     });
+    /// client.actions(mouse_actions).perform().await?;
+    /// ```
+    ///
+    /// See the documentation for [`Actions`] for more information.
+    pub fn actions(&mut self, actions: impl Into<ActionSequence>) -> Actions {
```
jonhoo, line 715: Ah, so, I proposed this split (`actions`/`perform`) only in relation to having a builder type. Since we abandoned that, I think this should now probably just be `perform_actions` :+1:

src/elements.rs
jonhoo: I see you renamed it to `css_value`, which seems reasonable :+1:

src/keys.rs
```diff
@@ -0,0 +1,263 @@
...
+        assert_eq!(Key::Control + "a", "\u{e009}a".to_string());
+        assert_eq!("a" + Key::Control, "a\u{e009}".to_string());
+    }
+
+    #[test]
+    fn test_key_string() {
+        assert_eq!(Key::Control + "a".to_string(), "\u{e009}a".to_string());
+        assert_eq!("a".to_string() + Key::Control, "a\u{e009}".to_string());
+    }
+
```
jonhoo, line 252: Let's also add a test for `Key + Key + String` (like Ctrl + Alt + e)

src/actions.rs
jonhoo: Wow, I must have been low on sleep. Yes, I meant `const`, in which case it should be `MOUSTE_BUTTON_LEFT`. I thought about a `#[repr(u64)]` enum, but I think that'll be annoying enough not to be worthwhile. And you already did it, so :+1:

src/actions.rs
```diff
@@ -0,0 +1,497 @@
...
+                },
+            )),
+            PointerAction::Cancel => {
+                WDActions::PointerActionItem::Pointer(WDActions::PointerAction::Cancel)
+            }
+        }
+    }
+}
+
+/// A sequence containing `Null` actions.
```
jonhoo, line 199: If you write:
```
[`NullAction`]
```
instead, it'll become a link in the documentation. You can, I think, even write:
```rust
/// A sequence containing [`Null` actions](NullAction).

Comment (jonhoo) [reply src/actions.rs:199]: We should probably do this on the other `*Actions` structs.

File: src/actions.rs [hunk 42]
```diff
@@ -0,0 +1,497 @@
...
+pub struct KeyActions {
+    /// An identifier to distinguish this sequence from others.
+    ///
+    /// Choose a meaningful string as it may be useful for debugging.
+    id: String,
+    /// The list of actions for this sequence.
+    actions: Vec<KeyAction>,
+}
+
+impl KeyActions {
```
jonhoo, line 242: Should we also implement `Default` for this and the other `*Actions` types with some default identifier, or is that unlikely to be useful/meaningful?

src/actions.rs
```diff
@@ -0,0 +1,497 @@
...
+    /// Create a new KeyActions sequence.
+    pub fn new(id: &str) -> Self {
+        Self {
+            id: id.to_string(),
```
jonhoo, line 246: The convention is generally to let the caller handle allocation, so to take `String` rather than `&str` if we actually need a `String` internally.

src/actions.rs
```diff
@@ -0,0 +1,497 @@
...
+    }
+}
+
+/// A sequence of actions to be performed.
+///
+/// See the documentation for [`Actions`] for more details.
+#[derive(Debug)]
+pub struct ActionSequence(pub(crate) WDActions::ActionSequence);
+
+/// Each sequence type implements `InputSource` which provides a `pause()` and a `then()`
```
jonhoo, line 374: Let's have a summary first line, something along the lines of:

> An source capable of providing inputs for a browser action chain.

And also link to the spec section on input sources.

src/actions.rs
```diff
@@ -0,0 +1,497 @@
+//! Actions functionality for WebDriver.
+use crate::elements::Element;
+#[cfg(doc)]
+use crate::keys::Key;
+use crate::Client;
+use std::fmt::Debug;
+use std::time::Duration;
+use webdriver::actions as WDActions;
+
+/// An action not associated with an input device (i.e. pause).
```
jonhoo, line 10:
```suggestion
/// An action not associated with an input device (e.g. pause).
```
`pause` is the only example we have currently, but I don't think we know that it's the only _possible_ null action for all future.

src/actions.rs
```diff
@@ -0,0 +1,497 @@
...
+    /// Pause action.
+    /// Useful for adding pauses between other key actions.
+    Pause {
+        /// The pause duration, given in milliseconds.
+        duration: Duration,
+    },
+    /// Pointer button down.
+    Down {
+        /// The mouse button index. Button values are as follows:
+        /// - Left = 0
```
jonhoo, line 111: Let's use the now-defined `const`s here for the example.

src/elements.rs
jonhoo: Actually, reading that, I _think_ we've actually been using the wrong term. I think in all these instances, these are "web element references" (as defined in the third item), not _the_ "web element identifier" (which is a constant string). I think we should quote the spec directly in the docs here:

> Each element has an associated web element reference that uniquely identifies the element across all browsing contexts. The web element reference for every element representing the same element must be the same. It must be a string, and should be the result of generating a UUID.

And also link to section 11.

jonhoo (changes requested):
I think we're getting really close! Left some more notes.

Re the move to `MoveBy` and friends, yeah, I think I like that better from an ergonomics perspective :+1:

src/actions.rs
stevepryde: If they don't need to be unique, then that's probably fine. I'll do some testing to find out.

src/elements.rs
stevepryde: I also renamed `ElementId` to `ElementRef`

src/client.rs
stevepryde: @jonhoo `Actions` is currently now a builder type anyway. 

Do we want:
```rust
client.actions(first_actions).and(second_actions).perform().await?
```

or revert to the original

```rust
client.perform_actions(actions).await?;
```

I prefer the second one, in the spirit of not trying to make this too high-level

src/actions.rs
stevepryde: Ok so they do need to be unique. In most cases you'd only have one of each device type but it's still possible to have multiple and in that case the id needs to be unique to the device.

I think we need the caller to specify the id.

src/client.rs
stevepryde: I've changed it to the second option. Let me know if you prefer something else.

src/actions.rs
jonhoo: Ah, good catch! In that case, let's also document that requirement.

src/client.rs
jonhoo: Ah, I see what you mean. Hmm... In some sense I didn't think of `and` as a _really_ being a builder compared to what we had before. I could go either way on this one, so whichever you prefer. I think that _if_ you go the first route, `fn actions` should take no arguments and return an _empty_ action sequence, rather than requiring the user to pass in the first one.

stevepryde: I think I'll stick with the second. Since the action chains could end up quite long and require multiple input sources etc, it feels better to separate the creation of the actions from the client method to perform them.

stevepryde:
Should `Element::css_value()` be called `Element::css_prop()` instead?

src/keys.rs
```diff
@@ -0,0 +1,278 @@
+//! Key codes for use with Actions.
+
+use std::fmt::{Display, Formatter};
+use std::ops::{Add, AddAssign};
+
+/// Key codes for use with Actions.
+#[derive(Debug, Clone)]
```
jonhoo, line 7: I think we should probably just also make this `Copy`.

src/keys.rs
```diff
@@ -1,9 +1,10 @@
 //! Key codes for use with Actions.
 
+use std::fmt::{Display, Formatter};
 use std::ops::{Add, AddAssign};
 
 /// Key codes for use with Actions.
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub enum Key {
```
jonhoo, line 8: Should this file now be called `key.rs`?

src/keys.rs
jonhoo: Then you can also avoid the `.clone()` in the `impl Display`.

src/actions.rs
```diff
@@ -107,28 +108,53 @@ pub enum PointerAction {
     },
     /// Pointer button down.
     Down {
-        /// The mouse button index. Button values are as follows:
-        /// - Left = 0
-        /// - Middle = 1
-        /// - Right = 2
+        /// The mouse button index.
+        ///
+        /// The following constants are provided:
```
jonhoo, line 113: Let's link to them with `[]`, and also add a note saying "but any mouse index can be used to represent the corresponding mouse button".

src/actions.rs
jonhoo: I don't think you ended up adding a link to the spec, unless I missed it?

tests/actions.rs
```diff
@@ -0,0 +1,235 @@
+//! Alert tests
+#[macro_use]
+extern crate serial_test;
+extern crate fantoccini;
```
jonhoo, line 4: This shouldn't be necessary?

tests/actions.rs
```diff
@@ -0,0 +1,235 @@
...
+    c.perform_actions(vec![key_pause].into()).await?;
+    assert!(now.elapsed().as_seconds_f64() >= 1.0);
+
+    // Test key down/up.
+    let mut elem = c.find(Locator::Id("text-input")).await?;
+    assert_eq!(elem.prop("value").await?.unwrap(), "");
+
+    let key_actions = KeyActions::new("key".to_string())
+        .then(KeyAction::Down { value: 'a' })
+        .then(KeyAction::Up { value: 'a' });
```
jonhoo, line 45: Is there a chance that this triggers key repetition and we get something like `aaa`?

tests/actions.rs
```diff
@@ -0,0 +1,235 @@
...
+}
+
+async fn actions_mouse(mut c: Client, port: u16) -> Result<(), error::CmdError> {
+    let sample_url = sample_page_url(port);
+    c.goto(&sample_url).await?;
+
+    // Test pause.
+    let mouse_pause = MouseActions::new("mouse".to_string()).pause(Duration::from_secs(1));
+    let now = Instant::now();
+    c.perform_actions(vec![mouse_pause].into()).await?;
```
jonhoo, line 60: With the change to `perform_actions`, I think this argument can just be `mouse_pause`, which'd be nice.

tests/actions.rs
```diff
@@ -0,0 +1,235 @@
...
+
+    elem.send_keys("fantoccini").await?;
+    assert_eq!(elem.prop("value").await?.unwrap(), "fantoccini");
+
+    // Test mouse MoveTo and MoveBy, by implementing drag-and-drop to select text
+    // in the text input element.
+    let mouse_actions = MouseActions::new("mouse".to_string())
+        // Move to the left edge of the input element.
+        // Offset by 1 pixel to ensure we click inside the element.
+        .then(PointerAction::MoveTo {
```
jonhoo, line 102: Won't this break if the text ends up wrapping? Do we know the text element will be sufficiently wide?

tests/elements.rs
```diff
@@ -0,0 +1,193 @@
...
+    assert_eq!(elem.css_value("display").await?, "none");
+    assert_eq!(elem.css_value("invalid-css-value").await?, "");
+    Ok(())
+}
+
+async fn element_tag_name(mut c: Client, port: u16) -> Result<(), error::CmdError> {
+    let sample_url = sample_page_url(port);
+    c.goto(&sample_url).await?;
+    let mut elem = c.find(Locator::Id("checkbox-option-1")).await?;
+    assert_eq!(elem.tag_name().await?, "input");
```
jonhoo, line 71: Should we `.to_lower_case()` here, or does the spec guarantee lower-case?

tests/local.rs
```diff
@@ -353,6 +352,74 @@ async fn resolve_execute_async_value(mut c: Client, port: u16) -> Result<(), err
...
+        Some(Duration::from_secs(60)),
+        Some(Duration::from_secs(60)),
+        Some(Duration::from_secs(30)),
+    );
+    c.set_timeouts(new_timeouts.clone()).await?;
+
+    let got_timeouts = c.get_timeouts().await?;
+    assert_eq!(got_timeouts, new_timeouts);
+
+    // Ensure partial update also works.
```
jonhoo, line 406: This behavior is surprising to me. I feel like it should be called `update_timeouts` rather than `set_timeouts`?

jonhoo (changes requested):
What's your reasoning for replacing `css_value` with `css_prop`?

src/actions.rs
stevepryde: Ah, I missed the link. Added now.

tests/actions.rs
stevepryde: Looks like the first two lines are needed but not the `extern crate fantoccini;`

tests/actions.rs
stevepryde: I don't actually know but I'd say "possibly". I guess it depends on the repeat delay in the keyboard settings on the underlying OS. In practice it seems fine. I've used key down/up sequences to mimic the behaviour of `send_keys()` before and it works fine.

In any case, I'll rework the test to remove this potential false positive.

tests/actions.rs
stevepryde: Since this is running in a fairly specific test environment I feel like it should be fine.
Just to be more sure I've added specific styling to the text box to enforce a specific width and font size.

tests/elements.rs
stevepryde: I wouldn't expect the spec to guarantee lower-case so I've added that.

tests/local.rs
stevepryde: The spec specifically allows you to specify one or more of the params, and `TimeoutConfiguration` has `skip_serializing_if = "Option::is_none"` on all fields.

I think `update_timeouts` does fit better so I've renamed it to that.

stevepryde:
> What's your reasoning for replacing `css_value` with `css_prop`?

I thought it was more consistent with `prop`. Technically they're all values, since we're getting the value of a property or a CSS property.

`attr` = Get the value of the specified attribute
`prop` = Get the value of the specified property
`css_prop` = Get the value of the specified CSS property

jonhoo:
> > What's your reasoning for replacing `css_value` with `css_prop`?
> 
> I thought it was more consistent with `prop`. Technically they're all values, since we're getting the value of a property or a CSS property.
> 
> `attr` = Get the value of the specified attribute `prop` = Get the value of the specified property `css_prop` = Get the value of the specified CSS property

It's a tough one — I honestly don't know. Your reasoning makes sense, but the spec also explicitly calls this "Get Element CSS Value", which is what made me gravitate towards `css_value`. And yet they also name the argument "property name" in the command, so I don't know what to think :p 

There's also the point that this returns the [_computed_ value](https://drafts.csswg.org/css-cascade-4/#computed-value), which isn't reflected in the name (and we should probably point out in the docs).

Really I suppose this is `computed_css_prop_value`, but that's an awful name. What does Selenium use?

In any case, I'm okay with `css_prop`.

jonhoo (changes requested):
Almost there! A few more bits.

stevepryde:
> > > What's your reasoning for replacing `css_value` with `css_prop`?
> > 
> > 
> > I thought it was more consistent with `prop`. Technically they're all values, since we're getting the value of a property or a CSS property.
> > `attr` = Get the value of the specified attribute `prop` = Get the value of the specified property `css_prop` = Get the value of the specified CSS property
> 
> It's a tough one — I honestly don't know. Your reasoning makes sense, but the spec also explicitly calls this "Get Element CSS Value", which is what made me gravitate towards `css_value`. And yet they also name the argument "property name" in the command, so I don't know what to think :p
> 
> There's also the point that this returns the [_computed_ value](https://drafts.csswg.org/css-cascade-4/#computed-value), which isn't reflected in the name (and we should probably point out in the docs).
> 
> Really I suppose this is `computed_css_prop_value`, but that's an awful name. What does Selenium use?
> 
> In any case, I'm okay with `css_prop`.

The `computed value` part makes sense. That's swaying me back towards `css_value`.
I think I'll leave the naming as is but mention that it's the `computed value` (with your link) in the docs

stevepryde:
> > > What's your reasoning for replacing `css_value` with `css_prop`?
> > 
> > 
> > I thought it was more consistent with `prop`. Technically they're all values, since we're getting the value of a property or a CSS property.
> > `attr` = Get the value of the specified attribute `prop` = Get the value of the specified property `css_prop` = Get the value of the specified CSS property
> 
> It's a tough one — I honestly don't know. Your reasoning makes sense, but the spec also explicitly calls this "Get Element CSS Value", which is what made me gravitate towards `css_value`. And yet they also name the argument "property name" in the command, so I don't know what to think :p
> 
> There's also the point that this returns the [_computed_ value](https://drafts.csswg.org/css-cascade-4/#computed-value), which isn't reflected in the name (and we should probably point out in the docs).
> 
> Really I suppose this is `computed_css_prop_value`, but that's an awful name. What does Selenium use?
> 
> In any case, I'm okay with `css_prop`.

I think the fact that it's the `computed value` makes me think it's better to leave it as `css_value`.
I've added the link to `computed value` in the docs.

Selenium-python calls it `value_of_css_property` which is a bit longer than I'd like. I think `css_value` is fine.

src/elements.rs
```diff
@@ -139,12 +238,30 @@ impl Element {
...
         match self.client.issue(cmd).await? {
             Json::String(v) => Ok(Some(v)),
+            Json::Bool(b) => Ok(Some(b.to_string())),
             Json::Null => Ok(None),
             v => Err(error::CmdError::NotW3C(v)),
         }
     }
 
-    /// Retrieve the text contents of this elment.
+    /// Look up the [`computed value`] of a CSS property for this element by name.
```
jonhoo, line 247:
```suggestion
    /// Look up the [computed value] of a CSS property for this element by name.
```

src/elements.rs
```diff
@@ -139,12 +238,30 @@ impl Element {
...
+    ///
+    /// `Ok(String::new())` is returned if the the given CSS property is not found.
+    ///
+    /// See [13.4 Get Element CSS Value](https://www.w3.org/TR/webdriver1/#get-element-css-value)
+    /// of the WebDriver standard.
+    ///
+    /// [`computed value`]: https://drafts.csswg.org/css-cascade-4/#computed-value
```
jonhoo, line 254:
```suggestion
    /// [computed value]: https://drafts.csswg.org/css-cascade-4/#computed-value
```

src/elements.rs
jonhoo: This isn't a code term

tests/elements.rs
```diff
@@ -1,6 +1,5 @@
 //! Element tests
-#[macro_use]
-extern crate serial_test;
+use serial_test::serial;
 extern crate fantoccini;
```
jonhoo, line 3: These can also go away now. Same in other files.

tests/actions.rs
```diff
@@ -1,6 +1,5 @@
 //! Alert tests
-#[macro_use]
-extern crate serial_test;
+use serial_test::serial;
```
jonhoo, line 2: I don't think there's a need for the empty line between this and the other `use` lines below any more.

tests/elements.rs
```diff
@@ -68,7 +67,7 @@ async fn element_tag_name(mut c: Client, port: u16) -> Result<(), error::CmdErro
     let sample_url = sample_page_url(port);
     c.goto(&sample_url).await?;
     let mut elem = c.find(Locator::Id("checkbox-option-1")).await?;
-    assert_eq!(elem.tag_name().await?.to_lowercase(), "input");
+    assert!(elem.tag_name().await?.eq_ignore_ascii_case("input"));
```
jonhoo, line 70: Bah, just realized that this won't tell us what the tag name _was_ if the test fails. Let's either leave it the way it was, or:
```suggestion
    let tag_name = elem.tag_name().await?;
    assert!(tag_name.eq_ignore_ascii_case("input"), "{} != input", tag_name);
```

jonhoo (changes requested):
Another few nits, but then we're there I think!

jonhoo (approved):
Excellent, thank you! Now we just need to figure out why CI fails. Looks like it's the `actions_mouse_move_test` that [fails](https://app.travis-ci.com/github/jonhoo/fantoccini/jobs/557596855#L902) with the error "move target out of bounds".

stevepryde:
> Excellent, thank you! Now we just need to figure out why CI fails. Looks like it's the `actions_mouse_move_test` that [fails on chrome](https://app.travis-ci.com/github/jonhoo/fantoccini/jobs/557596855#L902) with the error "move target out of bounds".

Hmm, it works on my PC. Wonder if it uses a different default viewport size? Would be good to make the window size explicit for testing if possible

stevepryde:
Also @jonhoo thank you for the review! I really appreciate the time and care taken on this 😄

jonhoo:
> > Excellent, thank you! Now we just need to figure out why CI fails. Looks like it's the `actions_mouse_move_test` that [fails on chrome](https://app.travis-ci.com/github/jonhoo/fantoccini/jobs/557596855#L902) with the error "move target out of bounds".
> 
> Hmm, it works on my PC. Wonder if it uses a different default viewport size? Would be good to make the window size explicit for testing if possible

This is why I was worried about the text area width in that other test :p
I honestly don't know if there's a way to set that. The tests all run in headless mode I believe, and `chromedriver` is started [here](https://github.com/jonhoo/fantoccini/blob/37d7b53522b238c09e24d210005a83ff05ef6a74/.travis.yml#L20).

jonhoo:
Hmm, looks like the same test still fails?

stevepryde:
> Hmm, looks like the same test still fails?

Yeah. I tried something different but same result. I wonder if simply setting the window rect beforehand is enough to solve it.

jonhoo:
Could be. I unfortunately don't think there is a way to tell the drivers to use windows of a particular size..

stevepryde:
Looks like the `actions_mouse_move` test is now passing. However the beta build is complaining about missing `#[must_use]`.

jonhoo:
Excellent! Yeah, seems like a new lint that we probably want to land fixes for (up for another quick PR?), but now we can finally land this at least!

jonhoo:
Thanks for all your hard work on this!

stevepryde:
> Excellent! Yeah, seems like a new lint that we probably want to land fixes for (up for another quick PR?), but now we can finally land this at least!

Yeah that should be a pretty easy one

c43721:
Thank you both for the time you spent and effort!

jonhoo:
Published as `0.19.0-rc.1` :tada:

Issue #153, opened 2022-02-06
Minimize instances where mutable access to self is required


stevepryde:
Currently it seems `fantoccini` is set up to require mutable access to `self` whenever issuing any webdriver command, even though this is strictly unnecessary in Rust terms.

I think it would be better if `fantoccini` did not try to "guess" the effect of an action on the state of the page in the browser, and instead just act as a minimal wrapper around the webdriver and leave any state management up to the caller/consumer.

Some examples of where I think the current mutability requirements feel awkward:
1. Things like requiring `mut self` when clicking an element just feel really heavy-handed especially in today's world where modern web apps often don't change the entire page in response to a single click.
2. Operations like `Element::is_displayed()` or `Client::get_window_rect()` require a mutable reference, which seems surprising, but it's currently required in order to call `Client::issue()`.
3. `Client::enter_frame()` takes `mut self` and returns `Result<Self, _>`. It's not clear to me what this is trying to prevent or enforce w.r.t. the calling code. Existing element references are not invalidated on the Rust side but will still return errors (at runtime) if you try to access them after switching to an iframe. This feels like it might provide a false sense of "correctness" while causing unnecessary friction. If the goal is to actually prevent access to elements outside the iframe at compile time after calling `enter_frame()`, the `mut self` won't do that.

The net effect of all of this is that the mutability requirements of `fantoccini` often feel arbitrary w.r.t. the actual web app and any operations on it. To get around this, calling code will probably just call `Client::clone()` anyway, which defeats the purpose.

If `fantoccini` instead took a minimalist approach, only requiring mutable access when Rust itself specifically required it, and allowing immutable access everywhere else, then I think it becomes much easier to reason about state changes in the browser as a result of calling code. It effectively removes `fantoccini` from the equation and allows the app itself to reason directly about the state changes in the browser as a result of any operations performed through `fantoccini`.

Happy to discuss any finer points here, either in the issue or in the PR.

PR #154, opened 2022-02-06
Make access to self immutable where possible

stevepryde:
This is intended firstly as a discussion point w.r.t. issue #153 and also to show how this would look with all of the mutability requirements minimized. 

The one method where I think capturing self might be warranted is `Client::close()`.

Perhaps its sig could be changed to:
```rust
pub async fn close(self) -> Result<Client, error::CmdError> {
```

Obviously all of this is a highly disruptive breaking change with many implications for calling code.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -91,7 +91,7 @@ impl Client {
     }
 
     /// Get the unique session ID assigned by the WebDriver server to this client.
-    pub async fn session_id(&mut self) -> Result<Option<String>, error::CmdError> {
+    pub async fn session_id(&self) -> Result<Option<String>, error::CmdError> {
         match self.issue(Cmd::GetSessionId).await? {
             Json::String(s) => Ok(Some(s)),
             Json::Null => Ok(None),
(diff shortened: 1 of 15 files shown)


stevepryde:
resolves #153

src/client.rs
```diff
@@ -812,10 +808,7 @@ impl Client {
     /// Screenshot](https://www.w3.org/TR/webdriver1/#dfn-take-element-screenshot) of the WebDriver
     /// standard.
     #[cfg_attr(docsrs, doc(alias = "Take Element Screenshot"))]
-    pub async fn screenshot_element(
-        &mut self,
-        element: Element,
-    ) -> Result<Vec<u8>, error::CmdError> {
+    pub async fn screenshot_element(&self, element: Element) -> Result<Vec<u8>, error::CmdError> {
```
jonhoo, line 811: Seeing this, I wonder if this should be a method on `Element` instead? Incidental to this PR though.

src/client.rs
```diff
@@ -885,7 +878,7 @@ impl Client {
             // TODO: get rid of this clone
             let current = current.clone();
             // TODO: and this one too
-            let mut c = c.clone();
+            let c = c.clone();
```
jonhoo, line 881: This clone _may_ no longer be necessary.

jonhoo (changes requested):
Part of the thinking for `&mut self` and requiring explicit clones was also to try to prevent users from shooting themselves in the foot by, say, having two threads writing into an element at the same time, but given the existence of `Clone` it's really wishful thinking anyway. So I'm on board with this change.

I like the idea of making `close` consume `self` — :+1: subject to that.

src/client.rs
stevepryde: Apparently it still is, due to lifetimes

src/client.rs
stevepryde: Probably. I'll move it (since we're already breaking everything 😅 )

jonhoo (approved):

PR #156, opened 2022-02-10
Move to GitHub actions

jonhoo:
We now have good coverage across OSes, code coverage checking, better
lint checking, feature combination checking, and msrv/minimal-versions
checks.


jonhoo:
@stevepryde Figured you'd probably find this interesting.

PR #160, opened 2022-02-20
Add WebDriverCompatibleCommands trait and implement it for Wcmd

### LINKED ISSUE #158: Add Trait for supporting other WebDriver-compatible commands (CDP, Selenium, Appium, etc)
Most (perhaps all) WebDriver commands simply consist of an endpoint, method, optional body, and JSON response, so this could be abstracted behind a trait. This would then allow other compatible commands to be supported in future, either as part of `fantoccini` (potentially CDP) or as part of downstream crates (`thirtyfour`, and potentially an appium crate later if anyone is interested).

The trait would look something like this:
```rust
pub trait WebDriverCompatibleCommand {
    fn endpoint(&self, session_id: Option<String>) -> Result<url::Url, url::ParseError>;
    fn method_and_body(&self) -> (http::Method, Option<serde_json::Value>);
}
```

That should be enough I think.

Since the `WebDriverCommand` enum would then become a subset, I'd recommend implementing the trait for it, and then changing the `Cmd::WebDriver(..)` variant to `WebDriver(Box<dyn WebDriverCompatibleCommand>)` instead of `WebDriver(Wcmd)`.

The other part that is required is to make `Client::issue()` completely public, or alternatively a
`pub fn issue_cmd(cmd: Box<dyn WebDriverCompatibleCmd>) -> Result<serde_json::Value, error::CmdError>`
to restrict it to just this particular subset of commands.

I think this is the last remaining change required to fully support the changeover for #145 :)
It also opens the door to add support for CDP, either natively in `fantoccini` or in a separate crate.

Note that actually adding support for any additional commands is not a requirement for resolving this issue.

stevepryde:
resolves #158

### DIFF
--- a/src/session.rs
+++ b/src/session.rs
@@ -322,11 +322,11 @@ impl Client {
     }
 
     /// Issue the specified [`WebDriverCompatibleCommand`] to the WebDriver instance.
-    pub async fn issue_cmd<T>(&self, cmd: T) -> Result<Json, error::CmdError>
-    where
-        T: Into<Box<dyn WebDriverCompatibleCommand + Send + 'static>>,
-    {
-        self.issue(Cmd::WebDriver(cmd.into())).await
+    pub async fn issue_cmd(
+        &self,
+        cmd: impl WebDriverCompatibleCommand + Send + 'static,
+    ) -> Result<Json, error::CmdError> {
+        self.issue(Cmd::WebDriver(Box::new(cmd))).await
     }
 
     pub(crate) fn is_legacy(&self) -> bool {
@@ -708,7 +708,7 @@ where
     /// [the spec]: https://www.w3.org/TR/webdriver/#list-of-endpoints
     fn issue_wd_cmd(
         &self,
-        cmd: Box<dyn WebDriverCompatibleCommand + Send + 'static>,
+        cmd: Box<impl WebDriverCompatibleCommand + Send + 'static + ?Sized>,
     ) -> impl Future<Output = Result<Json, error::CmdError>> {
         // TODO: make this an async fn
         // will take some doing as returned future must be independent of self
--- a/src/session.rs
+++ b/src/session.rs
@@ -39,7 +39,7 @@ pub(crate) enum Cmd {
 impl WebDriverCompatibleCommand for Wcmd {
     /// Helper for determining what URL endpoint to use for various requests.
     ///
-    /// This mapping is essentially that of https://www.w3.org/TR/webdriver/#list-of-endpoints.
+    /// This mapping is essentially that of <https://www.w3.org/TR/webdriver/#list-of-endpoints>.
     fn endpoint(
         &self,
         base_url: &url::Url,
--- a/src/session.rs
+++ b/src/session.rs
@@ -324,9 +324,9 @@ impl Client {
     /// Issue the specified [`WebDriverCompatibleCommand`] to the WebDriver instance.
     pub async fn issue_cmd<C>(&self, cmd: C) -> Result<Json, error::CmdError>
     where
-        C: WebDriverCompatibleCommand + Send + 'static,
+        C: Into<Box<dyn WebDriverCompatibleCommand + Send + 'static>>,
     {
-        self.issue(Cmd::WebDriver(Box::new(cmd))).await
+        self.issue(Cmd::WebDriver(cmd.into())).await
     }
 
     pub(crate) fn is_legacy(&self) -> bool {
@@ -707,12 +707,11 @@ where
     ///
     /// [the spec]: https://www.w3.org/TR/webdriver/#list-of-endpoints
     fn issue_wd_cmd(
-        &mut self,
-        cmd: Box<dyn WebDriverCompatibleCommand + Send>,
+        &self,
+        cmd: impl WebDriverCompatibleCommand + Send + 'static,
     ) -> impl Future<Output = Result<Json, error::CmdError>> {
         // TODO: make this an async fn
         // will take some doing as returned future must be independent of self
-
         let url = match cmd.endpoint(&self.wdb, self.session.as_deref()) {
             Ok(url) => url,
             Err(e) => return Either::Right(future::err(error::CmdError::from(e))),
--- a/src/session.rs
+++ b/src/session.rs
@@ -322,11 +322,11 @@ impl Client {
     }
 
     /// Issue the specified [`WebDriverCompatibleCommand`] to the WebDriver instance.
-    pub async fn issue_cmd(
-        &self,
-        cmd: Box<dyn WebDriverCompatibleCommand + Send>,
-    ) -> Result<Json, error::CmdError> {
-        self.issue(Cmd::WebDriver(cmd)).await
+    pub async fn issue_cmd<C>(&self, cmd: C) -> Result<Json, error::CmdError>
+    where
+        C: WebDriverCompatibleCommand + Send + 'static,
+    {
+        self.issue(Cmd::WebDriver(Box::new(cmd))).await
     }
 
     pub(crate) fn is_legacy(&self) -> bool {
@@ -677,7 +677,7 @@ where
 
                 // we're dealing with an implementation that only supports the legacy
                 // WebDriver protocol:
-                // https://github.com/SeleniumHQ/selenium/wiki/JsonWireProtocol
+                // https://www.selenium.dev/documentation/legacy/json_wire_protocol/
                 let session_config = webdriver::capabilities::LegacyNewSessionParameters {
                     desired: cap,
                     required: webdriver::capabilities::Capabilities::new(),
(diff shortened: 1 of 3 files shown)


src/session.rs
```diff
@@ -48,7 +294,7 @@ pub(crate) struct Task {
 }
 
 impl Client {
-    pub(crate) fn issue<C>(&self, cmd: C) -> impl Future<Output = Result<Json, error::CmdError>>
+    pub(crate) async fn issue<C>(&self, cmd: C) -> Result<Json, error::CmdError>
```
jonhoo, line 297: Hmm, I thought I did this on purpose so that the returned future wouldn't be tied to the lifetime of the borrow of `&self`, but if everything compiles fine I suppose it doesn't matter any more.

src/session.rs
```diff
@@ -59,22 +305,28 @@ impl Client {
...
+                io::ErrorKind::BrokenPipe,
+                "WebDriver session was closed while waiting",
+            )))
+        })
+    }
+
+    /// Issue the specified [`WebDriverCompatibleCommand`] to the WebDriver instance.
+    pub async fn issue_cmd(
+        &self,
+        cmd: Box<dyn WebDriverCompatibleCommand + Send>,
```
jonhoo, line 327: Should this take `impl Into<Box<dyn WebDriverCompatibleCommand + Send>>`? Or maybe just `C: WebDriverCompatibleCommand + Send`, and then we `Box` it internally?

src/session.rs
```diff
@@ -32,12 +33,257 @@ pub(crate) enum Cmd {
...
     },
-    WebDriver(Wcmd),
+    WebDriver(Box<dyn WebDriverCompatibleCommand + Send>),
+}
+
+impl WebDriverCompatibleCommand for Wcmd {
+    /// Helper for determining what URL endpoint to use for various requests.
+    ///
+    /// This mapping is essentially that of https://www.w3.org/TR/webdriver/#list-of-endpoints.
+    fn endpoint(
```
jonhoo, line 43: I assume there weren't any changes in the logic in this method compared to what we had before?

src/session.rs
```diff
@@ -32,12 +33,257 @@ pub(crate) enum Cmd {
...
+            WebDriverCommand::TakeElementScreenshot(ref we) => {
+                base.join(&format!("element/{}/screenshot", we.0))
+            }
+            WebDriverCommand::Print(..) => base.join("print"),
+            WebDriverCommand::Status => unreachable!(),
+            _ => unimplemented!(),
+        }
+    }
+
+    fn method_and_body(&self, request_url: &url::Url) -> (http::Method, Option<String>) {
```
jonhoo, line 146: And same here?

src/session.rs
```diff
@@ -563,136 +708,17 @@ where
     /// [the spec]: https://www.w3.org/TR/webdriver/#list-of-endpoints
     fn issue_wd_cmd(
         &mut self,
-        cmd: WebDriverCommand<webdriver::command::VoidWebDriverExtensionCommand>,
+        cmd: Box<dyn WebDriverCompatibleCommand + Send>,
```
jonhoo, line 711: This can definitely be generic. No need to force the boxing here.

src/wd.rs
```diff
@@ -7,9 +7,38 @@ use serde::{Deserialize, Serialize};
...
 use std::fmt;
+use std::fmt::Debug;
 use std::time::Duration;
 use webdriver::command::TimeoutsParameters;
 
+/// A command that can be sent to the WebDriver.
+///
+/// Anything that implements this command can be sent to [`Client::issue_cmd()`] in order
+/// to send custom commands to the WebDriver instance.
+pub trait WebDriverCompatibleCommand: Debug {
```
jonhoo, line 18: We should add a blanket `impl` of this for `&T` and `Box<T>` where `T: WebDriverCompatibleCommand`. Will make it a little nicer to work with.

src/wd.rs
```diff
@@ -7,9 +7,38 @@ use serde::{Deserialize, Serialize};
...
+    ///
+    /// The `url` will be the one returned from the `endpoint()` method above.
+    fn method_and_body(&self, request_url: &url::Url) -> (http::Method, Option<String>);
+
+    /// Return true if this command starts a new WebDriver session.
+    fn is_new_session(&self) -> bool {
+        false
+    }
+
+    /// Return true if this command is a legacy WebDriver command (prior to W3C).
```
jonhoo, line 36: Let's document a bit more what it means to return `true` from this.

jonhoo (changes requested):

src/session.rs
stevepryde: yeah I wasn't sure if I should leave it as is, but it did compile so I was like 🤷‍♂️ why not?

src/session.rs
stevepryde: I had to change the bit at the top to go from `self.wdb` to `base_url` but other than that it's identical

src/session.rs
stevepryde: Yep this one was a straight copy/paste except for renaming `url` => `request_url` (just wanted to disambiguate vs the `url` crate in the global namespace) and also `self.is_legacy` => `self.is_legacy()`.

src/session.rs
stevepryde: that's a good idea. Will do.

src/session.rs
stevepryde: Hmm, are you sure?

It is called from session:rs:492
```rust
                    Cmd::WebDriver(request) => {
                        // looks like the client setup is falling back to legacy params
                        // keep track of that for later
                        if request.is_legacy() {
                            self.is_legacy = true;
                        }

                        self.ongoing = Ongoing::WebDriver {
                            ack,
                            fut: Box::pin(self.issue_wd_cmd(request)),
                        };
                    }
```

the type of `request` here is `Box<dyn WebDriverCompatibleCommand + Send>`

src/session.rs
```diff
@@ -322,11 +322,11 @@ impl Client {
...
 
     /// Issue the specified [`WebDriverCompatibleCommand`] to the WebDriver instance.
-    pub async fn issue_cmd(
-        &self,
-        cmd: Box<dyn WebDriverCompatibleCommand + Send>,
-    ) -> Result<Json, error::CmdError> {
-        self.issue(Cmd::WebDriver(cmd)).await
+    pub async fn issue_cmd<C>(&self, cmd: C) -> Result<Json, error::CmdError>
+    where
+        C: WebDriverCompatibleCommand + Send + 'static,
```
jonhoo, line 327: Thinking some more about this, I _think_ we may want
```rust
C: Into<Box<dyn WebDriverCompatibleCommand + Send + 'static>>
```
so that if someone passes in a `Box<dyn WebDriverCompatibleCommand>`, we don't end up double-boxing it. I _think_ that should work. You _may_ need to write (if it's even legal):
```rust
impl From<Box<T>> for Box<dyn WebDriverCompatibleCommand + Send + 'static> where T: WebDriverCompatibleCommand + Send + 'static {
}
```

src/session.rs
jonhoo: Yeah, but then `impl WebDriverCompatibleCommand + Send + 'static` would work.

stevepryde: I'm having trouble getting this to work. I'm getting this:
```
error[E0277]: the size for values of type `dyn WebDriverCompatibleCommand + Send` cannot be known at compilation time
   --> src\session.rs:501:61
    |
501 | ...                   fut: Box::pin(self.issue_wd_cmd(request)),
    |                                          ------------ ^^^^^^^ doesn't have a size known at compile-time
    |                                          |
    |                                          required by a bound introduced by this call
    |
    = help: the trait `Sized` is not implemented for `dyn WebDriverCompatibleCommand + Send`
    = note: required because of the requirements on the impl of `WebDriverCompatibleCommand` for `Box<dyn WebDriverCompatibleCommand + Send>`
note: required by a bound in `session::Session::<C>::issue_wd_cmd`
   --> src\session.rs:711:19
    |
709 |     fn issue_wd_cmd(
    |        ------------ required by a bound in this
710 |         &self,
711 |         cmd: impl WebDriverCompatibleCommand + Send + 'static,
    |                   ^^^^^^^^^^^^^^^^^^^^^^^^^^ required by this bound in `session::Session::<C>::issue_wd_cmd`
```

src/session.rs
stevepryde: So if I do 
```rust
impl<T> From<Box<T>> ...
```

it works for Box<T> but not T.

The only way I found around it was to do 
```rust
impl<T> From<T> for Box<...>
```
but now we're back to double-boxing 🤷

stevepryde: I've simplified it for now - still boxing internally though.
I've added tests for this as well.

jonhoo (approved):
Looks good to me, thanks!

Issue #148, opened 2022-01-07
Don't silently hit a URL for raw client cookie support


LegNeato:
It is unexpected that using the raw client loads a completely different URL first to get cookies:

https://github.com/jonhoo/fantoccini/blob/6a7203e2307a397461ca633e0fd11719a88e339a/src/client.rs#L751

We should probably create typestates and specific functions to drive them and not do this automatically. Or perhaps let the URL be configured.

jonhoo:
Yeah, it's really awkward, but as best as I can tell from the WebDriver standard, it's the only way to get cookies for a _given_ domain. This is tracked for the spec in https://github.com/w3c/webdriver/issues/1238.

LegNeato:
My problem is the silent request is tripping bot detection, as this request + the "real" request happen too close together and the site sees me as a bot.

I'd like to at least be able to customize the url that is hit...would you take a PR to add that?

jonhoo:
Yeah, a PR seems like a good idea!

PR #162, opened 2022-02-23
Support controlling how cookies are fetched with raw clients

LegNeato:
Fixes https://github.com/jonhoo/fantoccini/issues/148.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -873,6 +873,8 @@ impl Client {
     /// Issue an HTTP request to the given `url` with all the same cookies as the current session.
     ///
     /// Calling this method is equivalent to calling `with_raw_client_for` with an empty closure.
+    ///
+    /// If greater control over cookie handling is desired, see `raw_cookie_client_for`.
     pub async fn raw_client_for(
         &self,
         method: Method,
@@ -887,6 +889,8 @@ impl Client {
     ///
     /// Before the HTTP request is issued, the given `before` closure will be called with a handle
     /// to the `Request` about to be sent.
+    ///
+    /// If greater control over cookie handling is desired, see `with_raw_cookie_client_for`.
     pub async fn with_raw_client_for<F>(
         &self,
         method: Method,
@@ -896,7 +900,6 @@ impl Client {
     where
         F: FnOnce(http::request::Builder) -> hyper::Request<hyper::Body>,
     {
-        let url = url.to_owned();
         // We need to do some trickiness here. GetCookies will only give us the cookies for the
         // *current* domain, whereas we want the cookies for `url`'s domain. So, we navigate to the
         // URL in question, fetch its cookies, and then navigate back. *Except* that we can't do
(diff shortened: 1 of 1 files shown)


src/client.rs
```diff
@@ -912,16 +915,64 @@ impl Client {
...
+    ) -> Result<hyper::Response<hyper::Body>, error::CmdError>
+    where
+        F: FnOnce(http::request::Builder) -> hyper::Request<hyper::Body>,
+    {
+        let old_url = self.current_url_().await?;
+        let url = old_url.clone().join(url)?;
+
+        if let Some(cookie_url) = cookie_url {
+            self.goto(cookie_url).await?;
+            self.back().await?;
```
jonhoo, line 965: Ah, this doesn't work. We need to gather the cookie _before_ we go back. That's the whole reason why the cookie URL is needed in the first place. `GetCookies` _only_ gives you the cookies for whatever the _current_ URL is, which is why we need to navigate to the target domain before calling it.

src/client.rs
```diff
@@ -912,16 +915,64 @@ impl Client {
...
+    }
+
+    /// Issue an HTTP request to the given `url` with all the same cookies as the current
+    /// session. If set, the client will first navigate to
+    /// `cookie_url` to retrieve cookies.
+    ///
+    /// Calling this method is equivalent to calling `with_raw_cookie_client_for` with an empty closure.
+    ///
+    /// `raw_client_for` is an alternative that automatically handles cookies.
+    pub async fn raw_cookie_client_for(
```
jonhoo, line 930: I'm not in love with the name here. I wonder if the way to go rather than having all of these different methods is to do a builder instead. I'm thinking:
```rust
impl Client {
    fn raw_request(&self) -> RawRequestBuilder;
}
impl RawRequestBuilder {
    fn method(&mut self, m: Method) -> &mut Self;
    fn url(&mut self, u: &str) -> &mut Self;
    fn cookie_url(&mut self, u: &str) -> &mut Self;
    fn map_request(&mut self, f: impl FnOnce(http::request::Builder) -> hyper::Request<hyper::Body>) -> &mut Self;
    fn send(self) -> Result<hyper::Response<hyper::Body>, error::CmdError>;
}
```
Given that we're going to do a breaking release anyway, fixing this properly now seems like a good idea. What do you think @stevepryde?

jonhoo (changes requested):
Thanks for taking the time to write up a PR! I left some notes.

src/client.rs
stevepryde: Yeah the builder idea is nice 👍 It should also allow for more capability than before (and future expansion) without adding more methods to `Client`.

I also think the original issue has merit as well - hitting a "random" url before the one specified by the caller does seem like surprising behaviour. I get why this is needed in order to get the cookies, but I think this needs to be more explicit. Something like the builder would make it more explicit in the calling code. The cookie fetch behaviour should also be documented as well (as it is in this PR currently).

Note that this could also allow the caller to NOT include cookies if they want (just skip the `cookie_url()`, thus avoiding the extra request) and then they have convenient access to a http client for any other use case where the cookies might be unnecessary or problematic. I'm not sure if this is useful or something you want to support. It's probably out of scope here at least.

LegNeato: Sweet, I'll convert this to a builder!

src/client.rs
LegNeato: Ah yeah, good point. Will fix.

src/client.rs
LegNeato: There are three cases here:

1. Don't hit a URL but send current cookies
2. Hit a URL and get cookies
3. Don't hit a URL and don't send current cookies

I think the builder should support all three.

jonhoo: I agree

jonhoo:
Gentle nudge on this @LegNeato

LegNeato:
Yep, sorry...I've been dealing with a move. I intend to get back to this in next week.

LegNeato:
Don't have time for this currently, closing until I do.

PR #163, opened 2022-03-04
Implement `Wheel` actions to support scrolling.

jaboatman:
The new webdriver standard includes `wheel` actions for simulating scroll wheels. https://www.w3.org/TR/webdriver/#wheel-actions

The upstream `webdriver` crate hasn't implemented it, but I worked around it here by implementing a new `WebDriverCompatibleCommand` that makes use of the new wheel implementation and the existing webdriver structures. This required a minimal amount of re-implementing structs from webdriver.

I'm not sure how universally the wheel APIs are implemented, but this is working just fine for me in `chromedriver 99.0.4844.51`.

This PR un-apologetically includes some other new methods that I needed around actions and the `Error` type. I can remove them if that is a blocker.

### DIFF
--- a/src/actions.rs
+++ b/src/actions.rs
@@ -241,12 +241,15 @@ impl NullActions {
 
 impl From<NullActions> for ActionSequence {
     fn from(na: NullActions) -> Self {
-        ActionSequence(WDActions::ActionSequence {
-            id: na.id,
-            actions: WDActions::ActionsType::Null {
-                actions: na.actions.into_iter().map(|x| x.into_item()).collect(),
-            },
-        })
+        ActionSequence(
+            WDActions::ActionSequence {
+                id: na.id,
+                actions: WDActions::ActionsType::Null {
+                    actions: na.actions.into_iter().map(|x| x.into_item()).collect(),
+                },
+            }
+            .into(),
+        )
     }
 }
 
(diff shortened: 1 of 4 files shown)


jonhoo:
Hmm, I'm torn on introducing additional APIs beyond what the `webdriver` crate implements. It adds more indirection, which makes the code harder to follow :thinking: What do you think @stevepryde ?

stevepryde:
I agree. It would be much nicer to get the wheel actions added to the `webdriver` crate. Not sure if that is feasible.

jonhoo:
It should be if it's a part of the spec! Should be feasible to send a PR to https://hg.mozilla.org/mozilla-central/file/tip/testing/webdriver through the Mozilla PR process.

jaboatman:
I agree, it should be part of webdriver. I haven't a clue how to make that happen, and I need this feature now, so I implemented it this way. If y'all want to figure out how to get it upstreamed to webdriver - even better. I don't currently have the time to figure that out though, so I'll continue on with my fork.

Thanks guys.

jonhoo:
The process is pretty similar to what it is on GitHub, just using Bugzilla instead. Basically you'd file an issue on there (see [this one](https://bugzilla.mozilla.org/show_bug.cgi?id=1710421) as an example). If you wanted to also do a PR, you follow [this process](https://firefox-source-docs.mozilla.org/contributing/contribution_quickref.html#firefox-contributors-quick-reference) :)

I'm going to go ahead and close this PR, and hopefully we'll eventually get this in upstream so you can avoid having to stay on your fork!

PR #174, opened 2022-06-30
Update webdriver to 0.46

complexspaces:
This PR updates `webdriver` to the latest version and makes the required adjustments for it to compile. I have not tested this locally since I'm unable to get `geckodriver` working, so I'm relying on CI to catch if some of these fields need filled out more.

Supersedes #173

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,7 @@ native-tls = ["hyper-tls"]
 rustls-tls = ["hyper-rustls"]
 
 [dependencies]
-webdriver = { version = "0.45", default-features = false }
+webdriver = { version = "0.46", default-features = false }
 url = "2.2.2"
 serde = { version = "1.0.103", features = ["derive"] }
 serde_json = "1.0.25"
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -23,7 +23,7 @@ native-tls = ["hyper-tls"]
 rustls-tls = ["hyper-rustls"]
 
 [dependencies]
-webdriver = { version = "0.45", default-features = false }
+webdriver = { version = "0.46", default-features = false }
```
jonhoo, line 26: Nice! I think this also means we can finally remove the `_time01` dependency a bit further down — would you mind doing that in this PR too?

complexspaces: Good call! I think that you're right. I removed this and the lockfile cleaned itself up quite a bit! I've pushed a commit handling this.

jonhoo (approved):

jonhoo:
Published in 0.19.1 :tada:

Issue #180, opened 2022-08-04
BrowserStack requires `charset=utf-8` in `Content-Type` header on Selenium >= 4 


lamalex:
Initiating a session on Selenium >= 4 requires the charset to be declared in the content type.
Although this feels like misuse on selenium's part, they seem to be dug in. See https://github.com/SeleniumHQ/selenium/issues/10913.

Setting `Content-Type: application/json; charset=utf-8` should not cause any breakage in existing implementations, but would expand compatibility.

jonhoo:
Oh interesting! That should be a straightforward enough modification here:
https://github.com/jonhoo/fantoccini/blob/e015d3763e6d07accea2fac13d6b9140436d8f69/src/session.rs#L746

I'm super swamped for time at the moment, but happy to approve a PR!

PR #183, opened 2022-08-11
Add `charset=utf-8` to session `Content-Type` header

lamalex:
Fixes: #180
BrowserStack (selenium v >=4) requires the charset to be set due to some
undisclosed security vulnerability, despite technically being implied by
the JSON spec.

Adding `charset=utf-8` should not have any backward compatibility issues
with current users.

### DIFF
--- a/src/session.rs
+++ b/src/session.rs
@@ -740,8 +740,12 @@ where
             );
         }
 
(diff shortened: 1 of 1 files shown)


jonhoo (approved):
Thanks!

jonhoo:
Released in 0.19.3 :tada:

Issue #185, opened 2022-08-25
Cannot access error variant information from webdriver::ErrorStatus enum


stevepryde:
It's pretty common when working with WebDriver to want to handle the specific errors returned, such as `ElementNotInteractable` etc.

I understand fantoccini wants to not expose the webdriver crate types etc but it would be good to provide an equivalent so that these specific error variants can be used in calling code.

There's a related issue where `StaleElementReference` is treated as `NoSuchElement` which is also not ideal, since they're not actually interchangeable and may warrant a different course of action depending on which error is returned.

If we don't want to expose the `ErrorStatus` enum, is it better to duplicate it and provide a `From` impl? Any better options?

jonhoo:
Unfortunately even a `From` impl would mean that updating webdriver would be a breaking change. I think it'd be okay for us to basically copy their error enumerations into this crate though, and then just mark those as `#[non_exhaustive]`!

PR #186, opened 2022-09-07
Add ErrorStatus enum and remove webdriver::error::WebDriverError

stevepryde:
- Add `ErrorStatus` enum
- Remove `webdriver::error::WebDriverError`
- Use `error::WebDriver` directly to avoid double-handling

NOTE: `StaleElementReference` is no longer combined with `NoSuchElement`.

I essentially copied the `ErrorStatus` enum from the webdriver crate, noting that the majority of the comments are public documentation from the WebDriver spec. Do we need to remove the comments that are not in the spec? I'm not sure what the MPL license requirements are.

Ideally I'd like to move the `NoSuchElement` etc. variants back into the `CmdError::Standard` variant - not sure how you feel about that. That would break almost every application using this crate :( 
Otherwise there may be confusion about those variants existing in `ErrorStatus` but they can never be captured that way.

This is already a breaking change, but the changes should be pretty minimal so far.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -113,24 +113,154 @@ pub enum CmdError {
 }
 
 impl CmdError {
-    /// Returns true if this error indicates that a matching element was not found.
-    ///
-    /// Equivalent to
-    /// ```no_run
-    /// # use fantoccini::error::{CmdError, ErrorStatus};
-    /// # let e = CmdError::NotJson(String::new());
-    /// let is_miss = if let CmdError::Standard(w) = e {
-    ///     matches!(w.error, ErrorStatus::NoSuchElement)
-    /// } else {
-    ///     false
-    /// };
-    /// ```
-    pub fn is_miss(&self) -> bool {
-        if let CmdError::Standard(w) = self {
-            matches!(w.error, ErrorStatus::NoSuchElement)
-        } else {
-            false
-        }
+    /// Return true if this error matches [`ErrorStatus::DetachedShadowRoot`].
+    pub fn is_detached_shadow_root(&self) -> bool {
+        matches!(self, CmdError::Standard(w) if w.error == ErrorStatus::DetachedShadowRoot)
+    }
+
+    /// Return true if this error matches [`ErrorStatus::ElementNotInteractable`].
+    pub fn is_element_not_interactable(&self) -> bool {
+        matches!(self, CmdError::Standard(w) if w.error == ErrorStatus::ElementNotInteractable)
+    }
+
+    /// Return true if this error matches [`ErrorStatus::ElementNotSelectable`].
+    pub fn is_element_not_selectable(&self) -> bool {
+        matches!(self, CmdError::Standard(w) if w.error == ErrorStatus::ElementNotSelectable)
+    }
+
+    /// Return true if this error matches [`ErrorStatus::InsecureCertificate`].
+    pub fn is_insecure_certificate(&self) -> bool {
+        matches!(self, CmdError::Standard(w) if w.error == ErrorStatus::InsecureCertificate)
+    }
+
+    /// Return true if this error matches [`ErrorStatus::InvalidArgument`].
+    pub fn is_invalid_argument(&self) -> bool {
+        matches!(self, CmdError::Standard(w) if w.error == ErrorStatus::InvalidArgument)
+    }
+
+    /// Return true if this error matches [`ErrorStatus::InvalidCookieDomain`].
+    pub fn is_invalid_cookie_domain(&self) -> bool {
+        matches!(self, CmdError::Standard(w) if w.error == ErrorStatus::InvalidCookieDomain)
+    }
+
+    /// Return true if this error matches [`ErrorStatus::InvalidCoordinates`].
+    pub fn is_invalid_coordinates(&self) -> bool {
+        matches!(self, CmdError::Standard(w) if w.error == ErrorStatus::InvalidCoordinates)
+    }
+
+    /// Return true if this error matches [`ErrorStatus::InvalidElementState`].
+    pub fn is_invalid_element_state(&self) -> bool {
+        matches!(self, CmdError::Standard(w) if w.error == ErrorStatus::InvalidElementState)
+    }
+
+    /// Return true if this error matches [`ErrorStatus::InvalidSelector`].
+    pub fn is_invalid_selector(&self) -> bool {
+        matches!(self, CmdError::Standard(w) if w.error == ErrorStatus::InvalidSelector)
+    }
+
+    /// Return true if this error matches [`ErrorStatus::InvalidSessionId`].
+    pub fn is_invalid_session_id(&self) -> bool {
+        matches!(self, CmdError::Standard(w) if w.error == ErrorStatus::InvalidSessionId)
+    }
+
+    /// Return true if this error matches [`ErrorStatus::JavascriptError`].
+    pub fn is_javascript_error(&self) -> bool {
+        matches!(self, CmdError::Standard(w) if w.error == ErrorStatus::JavascriptError)
+    }
+
+    /// Return true if this error matches [`ErrorStatus::MoveTargetOutOfBounds`].
+    pub fn is_move_target_out_of_bounds(&self) -> bool {
+        matches!(self, CmdError::Standard(w) if w.error == ErrorStatus::MoveTargetOutOfBounds)
+    }
+
+    /// Return true if this error matches [`ErrorStatus::NoSuchAlert`].
+    pub fn is_no_such_alert(&self) -> bool {
+        matches!(self, CmdError::Standard(w) if w.error == ErrorStatus::NoSuchAlert)
+    }
+
+    /// Return true if this error matches [`ErrorStatus::NoSuchCookie`].
+    pub fn is_no_such_cookie(&self) -> bool {
+        matches!(self, CmdError::Standard(w) if w.error == ErrorStatus::NoSuchCookie)
+    }
+
+    /// Return true if this error matches [`ErrorStatus::NoSuchElement`].
+    pub fn is_no_such_element(&self) -> bool {
+        matches!(self, CmdError::Standard(w) if w.error == ErrorStatus::NoSuchElement)
+    }
+
+    /// Return true if this error matches [`ErrorStatus::NoSuchFrame`].
+    pub fn is_no_such_frame(&self) -> bool {
+        matches!(self, CmdError::Standard(w) if w.error == ErrorStatus::NoSuchFrame)
+    }
+
+    /// Return true if this error matches [`ErrorStatus::NoSuchShadowRoot`].
+    pub fn is_no_such_shadow_root(&self) -> bool {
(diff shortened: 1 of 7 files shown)


stevepryde:
fixes #185

src/error.rs
```diff
@@ -284,19 +276,330 @@ impl From<InvalidWindowHandle> for CmdError {
     }
 }
 
+/// The error code returned from the WebDriver.
+#[derive(Debug, PartialEq)]
```
jonhoo, line 280: I think it's reasonable to derive `Hash` and `Eq` here as well.

src/error.rs
```diff
@@ -284,19 +276,330 @@ impl From<InvalidWindowHandle> for CmdError {
...
+    UnknownPath,
+
+    /// Indicates that a command that should have executed properly is not
+    /// currently supported.
+    UnsupportedOperation,
+}
+
+impl ErrorStatus {
+    /// Returns the string serialisation of the error type.
+    pub fn error_code(&self) -> &'static str {
```
jonhoo, line 427: It's weird to me that this is called error _code_, but returns a string. A code is (at least to me) generally a numerical representation. Do we even need this, or can we just implement `Display` instead?

jonhoo: I also think it's probably reasonable to implement `std::error::Error` for this type (which should be trivial since we derive `Debug` and if we implement `Display`).

src/error.rs
```diff
@@ -284,19 +276,330 @@ impl From<InvalidWindowHandle> for CmdError {
...
+impl Serialize for ErrorStatus {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        self.error_code().serialize(serializer)
+    }
+}
+
+impl From<webdriver::ErrorStatus> for ErrorStatus {
```
jonhoo, line 511: Unfortunately this `impl` isn't okay, as it'd still mean we're tied to the major version of the `webdriver` crate. This will have to be a private inherent method on `ErrorStatus` instead.

src/error.rs
```diff
@@ -284,19 +276,330 @@ impl From<InvalidWindowHandle> for CmdError {
...
+            UnknownCommand => ErrorStatus::UnknownCommand,
+            UnknownError => ErrorStatus::UnknownError,
+            UnknownMethod => ErrorStatus::UnknownMethod,
+            UnknownPath => ErrorStatus::UnknownPath,
+            UnsupportedOperation => ErrorStatus::UnsupportedOperation,
+        }
+    }
+}
+
+impl From<&str> for ErrorStatus {
```
jonhoo, line 550: I think `FromStr` (and possibly `TryFrom<&str>`) are the traits to implement instead. `From` should be infallible, and `UnknownError` feels like a cop-out.

src/error.rs
```diff
@@ -284,19 +276,330 @@ impl From<InvalidWindowHandle> for CmdError {
...
 pub struct WebDriver {
     /// Code of this error provided by WebDriver.
-    ///
-    /// Intentionally made private, so library users cannot match on it.
-    pub(crate) error: webdriver::ErrorStatus,
+    pub error: ErrorStatus,
 
     /// Description of this error provided by WebDriver.
-    pub message: Cow<'static, str>,
+    pub message: String,
```
jonhoo, line 594: Why not continue to allow this to be `&'static str`?

src/session.rs
```diff
@@ -872,80 +871,29 @@ where
...
-                                "received unknown error ({}) for METHOD_NOT_ALLOWED status code",
-                                error
-                            ),
-                        },
-                        _ => unreachable!("received unknown status code: {}", status),
-                    }
+                    body["error"]
+                        .as_str()
+                        .map(ErrorStatus::from)
+                        .unwrap_or(ErrorStatus::UnknownError)
```
jonhoo, line 877: This change means that we no longer require that the correct HTTP status code is used for each error kind. Is that intentional?

tests/local.rs
```diff
@@ -230,7 +230,7 @@ async fn stale_element(c: Client, port: u16) -> Result<(), error::CmdError> {
     .await?;
 
     match elem.click().await {
-        Err(error::CmdError::NoSuchElement(_)) => Ok(()),
+        Err(error::CmdError::Standard(_)) => Ok(()),
```
jonhoo, line 233: This change doesn't seem quite right?

jonhoo (changes requested):
Thanks for taking this on! I agree it's unfortunate the amount of breakage that may come about as a result of moving `NoSuchElement` (and the other `NoSuch*` variants), but I think it's the right call. They're standard responses, and should be handled as such :+1:

src/session.rs
stevepryde: Yes. Let me know if you disagree.
The string itself contains all of the information needed to parse the error type.

My reasoning is as follows:
1. It seems to me the only reason to be more strict with the HTTP status code is if we want to test the WebDriver implementation itself, but I doubt most users of this crate would be interested in that.
Anyone wanting to test the veracity of a particular WebDriver implementation should probably have their own test suite rather than relying on crates like this to report an incorrect HTTP status code.
2. As a user, if I happened to be using a WebDriver implementation that had a bug where it returned the correct error but incorrect HTTP status, I'd rather that my code continued to work than be blocked by an otherwise inconsequential upstream bug.

stevepryde: Also there is no situation where the same error string should be treated differently depending on the HTTP status code

tests/local.rs
stevepryde: hmm, yeah that's not specific enough after the change.

However, this leads to a new issue. Bringing `NoSuchElement` under `CmdError::Standard` creates more verbosity when checking for errors like this, which I suspect was the motivation for splitting these common variants out in the first place.

I'd like to solve this for all variants so that this kind of code (which is extremely common in browser automation) can be nicer to write.

Something like this is just too cumbersome:
```rust
match elem.click().await {
    Err(CmdError::Standard(w)) if w.error == ErrorStatus::NoSuchElement => {..}
    _ => {..}
}
```

One option is to do what the AWS SDK does and provider helper methods for checking the error variant. I.e.
```rust
match elem.click().await {
    Err(e) if e.is_no_such_element() => {..}
    _ => {..}
}
```
It's still not as nice as a flat enum though :(

Any other suggestions?

src/error.rs
stevepryde: I agree with the suggestion but it turns out this isn't actually used anywhere. Should be fine to remove it I think.

tests/local.rs
stevepryde: actually this particular test expects a stale element but the point remains the same

src/error.rs
stevepryde: Removed

src/session.rs
jonhoo: It's that very last sentence I'm concerned about, but given that that's not the case (and we could work around it if it were to be the case), I'm fine with the change :+1:

tests/local.rs
jonhoo: Yep, that was indeed the original motivation :) I think helpers on `Standard` is the way to go :+1:

src/error.rs
```diff
@@ -612,7 +560,7 @@ impl Error for WebDriver {}
 
 impl WebDriver {
     /// Create a new WebDriver error struct.
-    pub fn new(error: ErrorStatus, message: String) -> Self {
+    pub fn new(error: ErrorStatus, message: Cow<'static, str>) -> Self {
```
jonhoo, line 563: For convenience, I'd make this `impl Into<Cow<'static, str>>`

src/error.rs
```diff
@@ -499,58 +423,62 @@ impl ErrorStatus {
...
+
+impl Error for ErrorStatus {}
+
 impl Serialize for ErrorStatus {
     fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
     where
         S: Serializer,
     {
-        self.error_code().serialize(serializer)
+        self.to_string().serialize(serializer)
```
jonhoo, line 472: Given that all the strings are static, I think we should have a `ErrorStatus::description(&self) -> &'static str`, and then have both the `impl Display for ErrorStatus` and this code use that rather than going via `Display`

tests/local.rs
stevepryde: I've added helpers on `CmdError`

src/error.rs
```diff
@@ -113,24 +113,154 @@ pub enum CmdError {
...
-    pub fn is_miss(&self) -> bool {
-        if let CmdError::Standard(w) = self {
-            matches!(w.error, ErrorStatus::NoSuchElement)
-        } else {
-            false
-        }
+    /// Return true if this error matches [`ErrorStatus::DetachedShadowRoot`].
+    pub fn is_detached_shadow_root(&self) -> bool {
+        matches!(self, CmdError::Standard(w) if w.error == ErrorStatus::DetachedShadowRoot)
+    }
```
jonhoo, line 119: Not a blocker, but it'd be nice if we could turn all of these nearly-identical definitions into a macro. Something like:
```rust
macro_rules! is_helper {
    ($($variant:ident => $name:ident$(,)?),*) => {
        $(
            /// Return true if this error matches
            #[doc = concat!("[`ErrorStatus::", stringify!($variant), "`].")]
            pub fn $name(&self) -> bool {
                matches!(self, CmdError::Standard(w) if w.error == ErrorStatus::$variant)
            }
        )*
    }
}
```
You can then replace all of the methods with something like:
```rust
impl CmdError {
    is_helper! {
        Foo => is_foo,
        Bar => is_bar,
    }
}
```

src/error.rs
```diff
@@ -284,19 +367,322 @@ impl From<InvalidWindowHandle> for CmdError {
...
+    UnknownPath,
+
+    /// Indicates that a command that should have executed properly is not
+    /// currently supported.
+    UnsupportedOperation,
+}
+
+impl ErrorStatus {
+    /// Get the error string associated with this `ErrorStatus`.
+    pub fn description(&self) -> &'static str {
```
jonhoo, line 519: Not a blocker, but here too it might be nice to not have these string literals written out in two places in the code (here and in `FromStr`). May be worth turning them into constants, or writing a little helper macro to generate both this method and the `impl FromStr` at the same time?

jonhoo (approved):
Looks great now! I've left two comments that are more suggestions for improving the "niceness" of the code, but I wouldn't consider either a blocker. If you want me to merge this as-is, I'm happy to do that as well!

src/error.rs
stevepryde: Good idea. I actually watched your livestream of this 😄 This macro worked a treat. Thanks!

src/error.rs
stevepryde: I decided to do this as a macro as well.

jonhoo:
Released as 0.20.0-rc.1 :tada:

jonhoo:
I also now no longer find any reference to the `webdriver` crate in our docs output!

PR #191, opened 2022-10-15
Fix example basic run button locator

Kanan449:
This is a simple fix to resolve the "Run" button locator issue in the example file.


jonhoo (approved):
Nice, thanks!

PR #192, opened 2022-10-30
Fix Cookie -> WebDriverCookie same_site parsing

10ne1:
Fetching all cookies from fantoccini client and loading them back reveals that same_site is not properly parsed when converting from Cookie -> WebDriverCookie.

For example, client.get_all_cookies() can return both the following same_site values in different cookies, the second of which is transformed by the map call into an Option::None which gets ignored by WebDriverCookie serde skip_serializing_if = "Option::is_none":

same_site: Some(
        None,
    ),

same_site: None,

An example error when calling client.add_cookie() with the above "same_site: None":

thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Standard(WebDriver { error: InvalidArgument, message: "invalid argument: invalid 'sameSite'\n
(Session info: chrome=98.0.4758.80)"

This fixes the error by parsing the top-level None value to a None string as expected by WebDriverCookie.

### DIFF
--- a/src/cookies.rs
+++ b/src/cookies.rs
@@ -111,10 +111,13 @@ impl<'a> From<Cookie<'a>> for WebDriverCookie {
         let expiry = cookie
             .expires()
             .and_then(|e| e.datetime().map(|dt| dt.unix_timestamp() as u64));
-        let same_site = cookie.same_site().map(|x| match x {
-            SameSite::Strict => "Strict".to_string(),
-            SameSite::Lax => "Lax".to_string(),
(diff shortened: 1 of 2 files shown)


jonhoo (changes requested):
Oh, that's fascinating! I like the fix. Would you mind adding a test for this too?

jonhoo (approved):
Thanks!

jonhoo:
Released in `0.20.0-rc.2` :tada:

PR #193, opened 2022-11-02
Fully implement the print command

henrikhorluck:
Previously, the print command would default to using a GET, and not forwarding its parameters, but was otherwise implemented with the correct path.

I have tested that this does fix the issue, can add a test if you want as well


jonhoo (approved):
Nice catch, thanks!

jonhoo:
Published as `rc.3` :tada:

PR #208, opened 2023-03-20
feat: Improved error handling by using `thiserror`

OtaK:
No idea if you would like this to be in; Seems to simplify things a lot when it comes to implementing & extending errors. 

My branch is (very) out of date compared to the `main` branch. I'll get to rebasing it soon.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,20 +24,21 @@ rustls-tls = ["hyper-rustls"]
 
 [dependencies]
 webdriver = { version = "0.46", default-features = false }
-url = "2.2.2"
-serde = { version = "1.0.103", features = ["derive"] }
-serde_json = "1.0.25"
+url = "2.2"
+serde = { version = "1.0", features = ["derive"] }
+serde_json = "1.0"
 futures-core = "0.3"
 futures-util = "0.3"
 tokio = { version = "1", features = ["sync", "rt", "time"] }
 hyper = { version = "0.14", features = ["stream", "client", "http1"] }
-cookie = { version = "0.16.0", features = ["percent-encode"] }
+cookie = { version = "0.16", features = ["percent-encode"] }
 base64 = "0.13"
-hyper-rustls = { version = "0.23.0", optional = true }
-hyper-tls = { version = "0.5.0", optional = true }
-mime = "0.3.9"
+hyper-rustls = { version = "0.23", optional = true }
+hyper-tls = { version = "0.5", optional = true }
+mime = "0.3"
 http = "0.2"
 time = "0.3"
+thiserror = "1.0.37"
 
 [dev-dependencies]
 tokio = { version = "1", features = ["full"] }
--- a/src/error.rs
+++ b/src/error.rs
@@ -9,83 +9,61 @@ use std::str::FromStr;
 use url::ParseError;
 
 /// An error occurred while attempting to establish a session for a new `Client`.
-#[derive(Debug)]
+#[derive(Debug, thiserror::Error)]
 pub enum NewSessionError {
     /// The given WebDriver URL is invalid.
-    BadWebdriverUrl(ParseError),
+    #[error("webdriver url is invalid: {0}")]
+    BadWebdriverUrl(#[from] ParseError),
     /// The WebDriver server could not be reached.
-    Failed(HError),
+    #[error("webdriver server did not respond: {0}")]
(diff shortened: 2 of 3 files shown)


jonhoo:
Thanks! I'm not opposed to the idea, though I'm curious how much this adds to build time — could you measure that (`cargo build --timings`) and report?

OtaK:
Sure - I made sure to run `cargo clean` between executions. Seems there's a small compilation time gain (probably because the error handling impl is not inline compared to a handwritten impl)

### Main branch
[Cargo Build Timings — fantoccini 0.20.0-rc.4-main.pdf](https://github.com/jonhoo/fantoccini/files/11071706/Cargo.Build.Timings.fantoccini.0.20.0-rc.4-main.pdf)

### This branch
[Cargo Build Timings — fantoccini 0.20.0-rc.4-feat-error-handling.pdf](https://github.com/jonhoo/fantoccini/files/11071707/Cargo.Build.Timings.fantoccini.0.20.0-rc.4-feat-error-handling.pdf)

Cargo.toml
```diff
@@ -24,20 +24,21 @@ rustls-tls = ["hyper-rustls"]
 
 [dependencies]
 webdriver = { version = "0.46", default-features = false }
-url = "2.2.2"
-serde = { version = "1.0.103", features = ["derive"] }
-serde_json = "1.0.25"
+url = "2.2"
+serde = { version = "1.0", features = ["derive"] }
+serde_json = "1.0"
```
jonhoo, line 29: I'd like to avoid these changes — the minimal listed patch versions are often there for a reason too.

src/error.rs
```diff
@@ -9,83 +9,61 @@ use std::str::FromStr;
 use url::ParseError;
 
 /// An error occurred while attempting to establish a session for a new `Client`.
-#[derive(Debug)]
+#[derive(Debug, thiserror::Error)]
 pub enum NewSessionError {
     /// The given WebDriver URL is invalid.
-    BadWebdriverUrl(ParseError),
+    #[error("webdriver url is invalid: {0}")]
+    BadWebdriverUrl(#[from] ParseError),
```
jonhoo, line 16: Let's use `#[source]` instead of `#[from]` for any of these we don't already have an explicit `impl From` for. It's not always the case that a public `impl From` makes sense, _especially_ not a single one.

src/error.rs
```diff
@@ -9,83 +9,61 @@ use std::str::FromStr;
 use url::ParseError;
 
 /// An error occurred while attempting to establish a session for a new `Client`.
-#[derive(Debug)]
+#[derive(Debug, thiserror::Error)]
 pub enum NewSessionError {
     /// The given WebDriver URL is invalid.
-    BadWebdriverUrl(ParseError),
+    #[error("webdriver url is invalid: {0}")]
```
jonhoo, line 15: I think the general guidance from the error working group is that you _either_ print the inner error in the message _or_ you propagate it using `Error::source` (which both `#[from]` and `#[source]` do I believe). This currently does both. Let's remove `: {0}` here and stick with just exposing source.

src/error.rs
```diff
@@ -412,39 +310,38 @@ pub enum ErrorStatus {
...
-            Timeout => StatusCode::INTERNAL_SERVER_ERROR,
-            UnableToCaptureScreen => StatusCode::BAD_REQUEST,
-            UnableToSetCookie => StatusCode::INTERNAL_SERVER_ERROR,
-            UnexpectedAlertOpen => StatusCode::INTERNAL_SERVER_ERROR,
-            UnknownCommand => StatusCode::NOT_FOUND,
-            UnknownError => StatusCode::INTERNAL_SERVER_ERROR,
-            UnknownMethod => StatusCode::METHOD_NOT_ALLOWED,
-            UnknownPath => StatusCode::NOT_FOUND,
-            UnsupportedOperation => StatusCode::INTERNAL_SERVER_ERROR,
+            Self::ElementClickIntercepted
```
jonhoo, line 314: I'd prefer to make this change in another PR.

jonhoo:
Huh, that's surprising — usually extra proc macros add a decent amount to build time even just because `thiserror` would have to be built first. It could be because you're building with 32 cores, and so the added size of the dependency graph doesn't matter too much, though I think I'm fine merging a change like this anyway. Thanks for measuring :+1:

PR #212, opened 2023-04-02
[WIP] Ergonomic attribute macro for testing

jacfra:
This PR is an attempt to introduce an attribute macro for testing with fantoccini as a dev dependency.

Essentially:

```
#[fantoccini::test(chrome, firefox)]
async fn end_to_end_test(client: Client) -> Result<(), CmdError>{

}
```

Upon being run the test will duplicate the test logic and run it against both `chromedriver` and `geckodriver`. 

Ideally generating: 

```
running 2 tests
test end_to_end_test::chrome ... ok
test end_to_end_test::firefox ... ok
```

Notes:
- I borrowed from the existing (but currently private) `tester!` macro.

Areas Of Review Requested:
- I completly ignored the `rustls` feature and used `Clientbuilder::native()`. Why would someone use one vs the other?
- I removed the `session_id` `Arc<Mutex>` it wasn't clear to me that the code was making use of the `session_id` although perhaps that has something to do with my inexperience coding threads.

Future Features:
- Allow something like `"chrome:9515"` and `"firefox:4444"` and parse out the port numbers.


jonhoo:
I guess I'm not quite clear on how much value this really provides — it's not as though there's lots of boilerplate to setting up the tests, and it's pretty easy to do it through a regular function (see our own test suite). If anything, I think I'd want this to live in a separate crate in its own repo. I don't think it's something that needs to live in fantoccini itself.

PR #215, opened 2023-04-24
IE mode of the msedge works

### LINKED ISSUE #214: An error occurs in the IE mode of the Edge browser.
fantoccini is very useful library and I want to use with IE mode of the Edge browser. (I know I should not use Internet Explorer...)

## Expected Behavior
* The IE mode of the Edge browser works.

## Current Behavior
* The following code does not work well with the IE mode of the Edge browser.

```rust
use fantoccini::ClientBuilder;
use anyhow::Result;

#[tokio::main]
async fn main() -> Result<()> {
    let mut capabiliteies = serde_json::map::Map::new();
    capabiliteies.insert("browserName".into(), "internet explorer".into());
    let options = serde_json::json!({
        "ie.edgechromium": true
    });
    capabiliteies.insert("se:ieOptions".into(), options);
    let _client = ClientBuilder::native().capabilities(capabiliteies).connect("http://localhost:9515").await?;
    Ok(())
}
```

* error message
```
Error: webdriver did not create session: No matching capability sets found.
Unable to match capability set 0: goog:chromeOptions is an unknown extension capability for IE
```

## Possible Solution

* Node.js `selenium-webdriver` could works, and there is not `goog:chromeOptions`.
```js
Capabilities {
  map_: Map(11) {
    'acceptInsecureCerts' => false,
    'browserName' => 'internet explorer',
    'browserVersion' => '11',
    'pageLoadStrategy' => 'normal',
    'platformName' => 'windows',
    'proxy' => {},
    'se:ieOptions' => {
      browserAttachTimeout: 0,
      elementScrollBehavior: 0,
      enablePersistentHover: true,
      'ie.browserCommandLineSwitches': '',
      'ie.edgechromium': true,
      'ie.edgepath': '',
      'ie.ensureCleanSession': false,
      'ie.fileUploadDialogTimeout': 3000,
      'ie.forceCreateProcessApi': false,
      ignoreProtectedModeSettings: false,
      ignoreZoomSetting: true,
      initialBrowserUrl: 'http://localhost:52298/',
      nativeEvents: true,
      requireWindowFocus: false
    },
    'setWindowRect' => true,
    'strictFileInteractability' => false,
    'timeouts' => { implicit: 0, pageLoad: 300000, script: 30000 },
    'unhandledPromptBehavior' => 'dismiss and notify'
  }
}
```

* It works if I delete the following code in `session.rs`.
```rust
        // make chrome comply with w3c
        cap.entry("goog:chromeOptions".to_string())
            .or_insert_with(|| Json::Object(serde_json::Map::new()))
            .as_object_mut()
            .expect("goog:chromeOptions wasn't a JSON object")
            .insert("w3c".to_string(), Json::from(true));
```

## Steps to Reproduce

* run IEDriverServer

```
.\IEDriverServer.exe /port=9515 /log-level=DEBUG
```

* run script

```
cargo run
```

## Environment

* cargo.toml
```
[dependencies]
anyhow = "1.0.70"
fantoccini = { path = "./fantoccini" }
serde_json = "1.0.96"
tokio = { version = "1.27.0", features = ["full"] }
```
* OS: Windows11
* IE Webdriver: 4.8.1

mishan88:
Fixes #214

## Proposed Changes

- Remove adding `goog:chromeoptions` to capabilities

### DIFF
--- a/src/session.rs
+++ b/src/session.rs
@@ -630,6 +630,15 @@ where
             cap.insert("pageLoadStrategy".to_string(), Json::from("normal"));
         }
 
+        // make chrome comply with w3c
+        if cap.get("browserName") == Some(&Json::from("chrome")) {
+            cap.entry("goog:chromeOptions".to_string())
+                .or_insert_with(|| Json::Object(serde_json::Map::new()))
+                .as_object_mut()
+                .expect("goog:chromeOptions wasn't a JSON object")
+                .insert("w3c".to_string(), Json::from(true));
+        }
+
         let session_config = webdriver::capabilities::SpecNewSessionParameters {
             alwaysMatch: cap.clone(),
             firstMatch: vec![webdriver::capabilities::Capabilities::new()],
--- a/src/session.rs
+++ b/src/session.rs
@@ -631,7 +631,7 @@ where
         }
 
         // make chrome comply with w3c
-        if cap.get("browserName") == Some(&Json::from("chrome")) {
+        if cap.get("browserName") != Some(&Json::from("internet explorer")) {
             cap.entry("goog:chromeOptions".to_string())
                 .or_insert_with(|| Json::Object(serde_json::Map::new()))
                 .as_object_mut()
--- a/src/session.rs
+++ b/src/session.rs
@@ -630,13 +630,6 @@ where
             cap.insert("pageLoadStrategy".to_string(), Json::from("normal"));
         }
 
-        // make chrome comply with w3c
-        cap.entry("goog:chromeOptions".to_string())
-            .or_insert_with(|| Json::Object(serde_json::Map::new()))
-            .as_object_mut()
-            .expect("goog:chromeOptions wasn't a JSON object")
-            .insert("w3c".to_string(), Json::from(true));
-
         let session_config = webdriver::capabilities::SpecNewSessionParameters {
             alwaysMatch: cap.clone(),
             firstMatch: vec![webdriver::capabilities::Capabilities::new()],


jonhoo:
Ah, so, I don't actually want to _remove_ this block — it's needed for Google Chrome support. Instead, I want to just not run the block if `se:ieOptions` is set in `caps` (so just place the block you removed inside an `if` instead).

mishan88:
Sorry. I misunderstood your intention. I'll rewrite later.

mishan88:
I rewrote. If capabilities' `browserName` is chrome, add `goog:chromeOptions`.

src/session.rs
```diff
@@ -631,11 +631,13 @@ where
         }
 
         // make chrome comply with w3c
-        cap.entry("goog:chromeOptions".to_string())
-            .or_insert_with(|| Json::Object(serde_json::Map::new()))
-            .as_object_mut()
-            .expect("goog:chromeOptions wasn't a JSON object")
-            .insert("w3c".to_string(), Json::from(true));
+        if cap.get("browserName") == Some(&Json::from("chrome")) {
```
jonhoo, line 634: Since no other browser has a problem with this, and not everyone passes in `browserName` as a capability, I'd like to instead special-case Internet Explorer here. Only if the browser name specifically is internet explorer should we _not_ include this. Makes sense?

jonhoo:
Excellent, thank you! I left a comment.

src/session.rs
mishan88: I agree with you.
I think `goog:chromeOptions` is only for chrome, but only IE driver does not accept this capability.

mishan88: I’m sorry for the late comment. I fixed it.

jonhoo (approved):

jonhoo:
Thank you!

PR #216, opened 2023-04-27
Implement wheel support.

jaboatman:


src/actions.rs
```diff
@@ -404,6 +404,100 @@ impl From<TouchActions> for ActionSequence {
...
+        })
+    }
+}
+
+/// An action performed with a wheel device.
+///
+/// See [15.4.4 Wheel Actions](https://www.w3.org/TR/webdriver/#wheel-actions) of the
+/// WebDriver standard.
+#[derive(Debug, Clone)]
+pub enum WheelAction {
```
jonhoo, line 451: Let's add the [`#[non_exhaustive]`](https://doc.rust-lang.org/reference/attributes/type_system.html#the-non_exhaustive-attribute) attribute to this so that we can add additional variants (like maybe `ScrollFromAbsolute` to have a different `origin`) in the future without breaking backwards compatibility.

src/actions.rs
```diff
@@ -404,6 +404,100 @@ impl From<TouchActions> for ActionSequence {
...
+    /// Pause action.
+    /// Useful for adding pauses between other key actions.
```
jonhoo, line 453: nit: Add an extra newline between these so that the first line is treated as a summary that is shown in the containing module type listing. Otherwise, this will be treated as all being a single paragraph.

jonhoo (changes requested):
Sorry for the very late review to this! It mostly looks good, just some minor comments. Main thing is leaving a space for us to potentially add a way to use a different scroll origin in the future.

src/actions.rs
jaboatman: Done.

jaboatman:
> Sorry for the very late review to this! It mostly looks good, just some minor comments. Main thing is leaving a space for us to potentially add a way to use a different scroll origin in the future.

No problem, I had forgotten all about it too

I updated it per your comments.

jonhoo:
Ah, I think you may need to merge from `main` to get CI passing.

jonhoo:
Heh, you may need to merge `main` _again_ actually to get the test improvements from #228.

jonhoo:
https://github.com/jonhoo/fantoccini/pull/231 will also take care of the `webdriver` bump.

jonhoo:
Released as rc.6 :tada:

PR #218, opened 2023-05-17
Update hyper-rustls to 0.24

complexspaces:
This PR is similar to https://github.com/jonhoo/fantoccini/pull/142 and updates `hyper-rustls` to the latest version to stay in sync with the rest of the `rustls` ecosystem.

Additionally, I made a small change to how the `openssl` dependency was being specified. As it was, it was forcing any user of the  `fantoccini` crate to depend on `openssl` even if you weren't using the `native-tls` feature.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -309,9 +309,9 @@ dependencies = [
 
 [[package]]
 name = "hyper-rustls"
-version = "0.23.0"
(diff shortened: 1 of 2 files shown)


jonhoo (approved):
Nice, thanks!

Issue #195, opened 2022-11-08
`Session::with_capabilities_and_connector` hides the WebDriver response


OtaK:
Hi!

I'm currently implementing a WASM browser-only test runner based on WebDriver & fantoccini and I was thinking of using WebDriver BiDi protocol to follow progress on tests. 

But following up on that I couldn't find a way to get the original NewSession command response which contains the `webSocketUrl` that webdriver creates for the communication - after providding the `{ "websocketUrl": true, ...}` capability.

https://w3c.github.io/webdriver-bidi/#establishing

Would it be possible to save the session handshake response somewhere and access it later?

Note: If needed I can make the PR myself. I honestly don't think WebDriver BiDi support is needed *in* fantoccini right now as it's still a W3C draft - but nonetheless at least partially implemented in all webdriver implementations -. And it would pull in a couple of other dependencies that are websocket-related. But if we could just access the NewSession response and extract the websocket URL, we could establish the connection from the outside quite easily for those use-cases.

Thanks!

jonhoo:
That sounds like an entirely reasonable thing to add! I'm a fan of storing the NewSession response and making it accessible later :+1: A PR would be very welcome.

PR #223, opened 2023-07-21
feat: expose the returned remote capabilities on the Client

RomFouq:
This would  tackle #195 by storing the remote capabilities (obtained with the new session response) in the `Client`, and exposing them via a public getter for later use.
This internally leverages the already existing `webdriver::response::NewSessionResponse` to map the response.

Being able to access the remote capabilities is necessary to be able to obtain the `webSocketUrl` remote capability and [establish a WebDriver BiDi connection](https://w3c.github.io/webdriver-bidi/#establishing).

Notes:
- The remote capabilities are stored in and exposed on the `Client` because it was the easiest, not sure if it would make more sense to have them on the `Session` instead.
- The public getter is named `.remote_capabilities()`, as it exposes the capabilities returned by the remote end, maybe you would prefer another name as I don't think `fantoccini` uses the local end/remote end terminology from the W3C specification.
- I've replaced the `.remove()`/`.insert()`-back-in-case-of-error JSON value extraction with `.get()`/`.to_owned()` as we now also need to access capabilities as well; do you think it would be better to avoid this clone?


src/client.rs
```diff
@@ -114,6 +115,11 @@ impl Client {
         }
     }
 
+    /// Get the capabilities returned by the remote end when opening the session.
+    pub fn remote_capabilities(&self) -> &Capabilities {
```
jonhoo, line 119: I think I'm okay with this just being called `capabilities` actually. I doubt anyone would expect that this is "the capabilities we originally sent to the server". And we can also clarify this in the doc text. What do you think?

src/session.rs
```diff
@@ -648,9 +654,10 @@ where
             .map(Self::map_handshake_response)
             .await
         {
-            Ok(_) => Ok(Client {
+            Ok(NewSessionResponse { capabilities, .. }) => Ok(Client {
                 tx,
                 is_legacy: false,
+                remote_capabilities: capabilities.as_object().unwrap().clone(),
```
jonhoo, line 660: Seeing this, should we perhaps just store (and expose) the entire `NewSessionResponse` rather than _only_ the capabilities? There's not much else in there at the moment, but the session id may be handy, and who knows if more fields may be added in the future?

jonhoo (approved):
This is a great addition, thank you! Left a few questions, but none of them major.

src/client.rs
```diff
@@ -114,6 +117,20 @@ impl Client {
         }
     }
 
+    /// Get the response obtained when opening the session.
+    ///
+    /// Returns `None` if no session has yet been opened.
+    pub fn session_creation_response(&self) -> Option<&NewSessionResponse> {
```
jonhoo, line 123: Ah, we can't actually directly return the type from `webdriver` here, as that would mean upgrading `webdriver` to a new major version would also require a new major version of `fantoccini`. This is what the [`wd` module](https://github.com/jonhoo/fantoccini/blob/main/src/wd.rs) is for — in there we have our own replicas of the `webdriver` types that are under _our_ control, and that we _can_ return. There, we can _also_ `#[derive(Clone)]` so that you won't need the `Arc` any more :tada: 

Just a note for when you add that type though, make sure to mark it as `#[non_exhaustive]` so that we can add fields to it in the future if the `webdriver` crate does!

jonhoo (changes requested):

src/wd.rs
```diff
@@ -345,3 +345,33 @@ impl TimeoutConfiguration {
...
+        &self.session_id
+    }
+
+    /// Get the remote end capabilities.
+    pub fn capabilities(&self) -> Option<&Capabilities> {
+        self.capabilities.as_ref()
+    }
+}
+
+impl From<webdriver::response::NewSessionResponse> for NewSessionResponse {
```
jonhoo, line 370: Unfortunately, even this `impl` isn't okay — it still makes `webdriver` part of the public API. Instead what you'll need is an
```rust
impl NewSessionResponse {
  fn from_wd(nsr: webdriver::response::NewSessionResponse) -> Self
```
so that there are no public `impl`s that name `webdriver`.

jonhoo (changes requested):
Nice! Very nearly there :tada:

jonhoo:
Also, you may want to merge from `main` to get a bunch of test fixes.

jonhoo (approved):
Thanks!

jonhoo:
Totally fine to bump the minimum version of `serde_json`!

jonhoo (approved):

jonhoo:
I'd generally prefer only bumping it to what we actually need (what goes in Cargo.toml is a *minimum* version after all). But not the end of the world if it preserves MSRV

jonhoo (approved):
Thank you!

jonhoo:
Released in v0.20.0-rc.7 :tada:

PR #226, opened 2023-08-13
Wikipedia changed again

jonhoo:
Someone please turn this test local


jonhoo:
The errors are transient wikipedia 503s, so we're good.

PR #228, opened 2023-08-15
Replaced Wikipedia links with local tests

ChaseCares:
Hello, this PR aims to replace the Wikipedia links with local tests. 12 of the tests have been converted without issue and currently pass on Firefox and Chrome. I broke this into two commits, the first demonstrates the changes and the second moves the tests to local.rs. Hopefully this improves clarity and makes it easier to merge.

There are two tests that I would like input on:

`wait_for_navigation_test()` Redirect from a file to Wikipedia, this is a bit awkward with a local URL. The URL after a local redirection may be dependent on where the test was performed. My proposed solution is to redirect to `sample_page.html` in `redirect_test.html` and change [254](https://github.com/jonhoo/fantoccini/compare/main...ChaseCares:remove_wikipedia_links?expand=1#diff-4a82577c1b3354b0e61c53badadb94977e59ccae42aad1b762c14a68517abbd2R254) to `assert!(url.to_string().ends_with("sample_page.html"));`. I don't love this solution, though. Is there a other method that would be preferable?

The other test is [handle_cookies_test()](https://github.com/jonhoo/fantoccini/compare/main...ChaseCares:remove_wikipedia_links?expand=1#diff-4a82577c1b3354b0e61c53badadb94977e59ccae42aad1b762c14a68517abbd2R269-R270). The issue arises with [cookie.set_domain()](https://github.com/jonhoo/fantoccini/compare/main...ChaseCares:remove_wikipedia_links?expand=1#diff-4a82577c1b3354b0e61c53badadb94977e59ccae42aad1b762c14a68517abbd2R269-R270), Firefox will allow you to forgo this parameter, Chrome however, throws an error. I tried most of the suggestions from [SO](https://stackoverflow.com/questions/1134290/cookies-on-localhost-with-explicit-domain) to no avail. ~~I will continue to search for a solution~~ This does not seem possible on Chrome, [reasoning](https://bugs.chromium.org/p/chromium/issues/detail?id=56211#c1).

Potential options:

1. Leave the test as external for both Chrome and Firefox
2. Split the test, by creating a local only test for Firefox and an external test for Chrome

Edit: I reverted the changes to the two above functions from my original commit. I will leave these until further guidance is given.

Code change justifications:
[R 58](https://github.com/jonhoo/fantoccini/commit/cd6b4bdcdb4cf34c772033722cea57ff3c99e56e#diff-4a82577c1b3354b0e61c53badadb94977e59ccae42aad1b762c14a68517abbd2R58)
With Wikipedia the form is submitted we check for `.../wiki/Foobar`, I changed this to `...?search=foobar`, I believe this still serves the same function, while also handling a 304 redirect.

[R 125](https://github.com/jonhoo/fantoccini/commit/cd6b4bdcdb4cf34c772033722cea57ff3c99e56e#diff-4a82577c1b3354b0e61c53badadb94977e59ccae42aad1b762c14a68517abbd2R126)
I added `globe.html` to act as the source for the image tag, this could potentially be another page or a proper image, whatever is preferred. The assert could be changed to exact bates, rather than not empty, this may be more precise.

[L 208](https://github.com/jonhoo/fantoccini/compare/main...ChaseCares:remove_wikipedia_links?expand=1#diff-4a82577c1b3354b0e61c53badadb94977e59ccae42aad1b762c14a68517abbd2L208-L213)
I removed this assert since the page is static and the hidden "edit preview settings" shouldn't be an issue.

let me know if there is anything else I can do to improve this PR.
Chase


ChaseCares:
Additionally, while working with the tests I noticed that there are some redundant test cases and I believe I could overall improve the test coverage. Would a PR for this be something the project would be interested in?

jonhoo (approved):
This looks great, thank you!
As for the two questions, I think it's totally fine to switch the redirect test to use `ends_with`. It's not _beautiful_, but it still ends up testing what we care about. And let's just keep the cookie test remote for now — we may just not be able to get around that.

jonhoo:
I'll go ahead and merge this as-is, and then we can do the `wait_for_navigation` test change in a separate mini-PR!

ZoeS17:
For reference I took the liberty of double checking and it seems Wikipedia is marked `Creative Commons Attribution Share-Alike` [Link to highlight](https://en.wikipedia.org/wiki/Main_Page#:~:text=Text%20is%20available%20under%20the%20Creative%20Commons%20Attribution%2DShareAlike%20License%204.0)

jonhoo:
Thanks! Doesn't matter in this case though since this PR doesn't replicate any of the Wikipedia stuff 😊

PR #229, opened 2023-08-15
Localize tests resolves #76

### LINKED ISSUE #76: Change tests to use local HTML files
#75 added the necessary infrastructure for running tests against locally hosted HTML files rather than relying on wikipedia.org staying static. We should move pretty much all the tests to use this infrastructure so that the test suite can be run without an internet connection, and so that we do not run the risk of the tests failing if the relevant Wikipedia HTML should change.

brandonsimpson21:
localizes `finds_all_inner`, `finds_sub_elements`, `window_size_inner`, `window_position_inner`, and `window_rect_inner` tests.


jonhoo:
Thanks for taking this on! I'm going to semi-randomly go with #228, partially because it came in slightly earlier, and partially because they've also trimmed down the HTML to not actually hold the true Wikipedia HTML (which we don't really need for tests).

PR #232, opened 2023-08-27
Migrate wait_for_nav test to use local files

ChaseCares:
This PR removes Wikipedia as a dependency for the `wait_for_navigation` test, it also moves `redirect_test.html` in with the other HTML test documents.

Thanks.
Chase


jonhoo (approved):
Excellent, thank you!

PR #240, opened 2024-01-26
examples don't seem to use panic! or unwrap anymore

maxcountryman:
This note seems outdated so here it's removed.


jonhoo (approved):
Ah, good catch, thank you!

PR #245, opened 2024-02-17
Bump to hyper 1.

jonhoo:
I'm not super happy about the BoxBody stuff, since many of the involved methods would happily work without boxing if the body generic type was on `Client::request` rather than on `Client` itself. But alas.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -89,9 +89,9 @@ dependencies = [
 
 [[package]]
 name = "cookie"
-version = "0.17.0"
+version = "0.18.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7efb37c3e1ccb1ff97164ad95ac1606e8ccd35b3fa0a7d99a304c7f4a428cc24"
+checksum = "3cd91cf61412820176e137621345ee43b3f4423e589e7ae4e50d601d93e35ef8"
 dependencies = [
  "percent-encoding",
  "time",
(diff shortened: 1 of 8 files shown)


jonhoo:
@stevepryde Do you think it's worth cutting a non-rc release of 0.20, then landing this, and then cutting the first rc of 0.21 immediately after? Just because of how pervasive the hyper-1 bump is across the ecosystem?

stevepryde:
> @stevepryde Do you think it's worth cutting a non-rc release of 0.20, then landing this, and then cutting the first rc of 0.21 immediately after? Just because of how pervasive the hyper-1 bump is across the ecosystem?

I think this makes sense.

However (unrelated to this), I've recently decided to take `thirtyfour` in a different (and more opinionated) direction and as a result it no longer depends on `fantoccini`. The decision was purely technical, relating to my underlying goals for the project around test automation. Not that I have a lot of time to work on it at the moment 😫 but I hope to get there eventually.

jonhoo:
@stevepryde All good, thanks for the input and the help over the years at least, and best of luck!

jonhoo:
0.20.0 without this change has now been cut :tada: 
https://github.com/jonhoo/fantoccini/pull/260

jonhoo:
Chrome failures on Windows are now due to https://issues.chromium.org/issues/42323434

jonhoo:
Chrome on Windows is still flakey/hangy on Windows, but about the same as on `main`, so going to land this now.

jonhoo:
Released as 0.21.0 :tada:

PR #248, opened 2024-04-27
Correct `set_window_size` docs

Carlosted:
Corrects a slight imprecision in the docs of said function :)


jonhoo (approved):
Good catch, thanks!

PR #257, opened 2024-06-27
feat(client): Implement `RawRequestBuilder` for raw HTTP requests

### LINKED ISSUE #148: Don't silently hit a URL for raw client cookie support

surajk-m:
Closes #148 

This PR introduces a `RawRequestBuilder` struct to handle raw HTTP requests with optional cookie handling.

_Builder Pattern Implementation:_

- `RawRequestBuilder` for setting HTTP method `(method)`, URL `(url)`, optional cookie URL `(cookie_url)`, request modification function `(map_request)`, and sending the request `(send)`.
-  Added `raw_request` method to the Client struct to initialize the `RawRequestBuilder`.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -185,7 +185,7 @@ pub struct RawRequestBuilder<'a> {
     method: Method,
     url: String,
     cookie_url: Option<String>,
-    request_modifier: Option<Box<dyn FnOnce(http::request::Builder) -> hyper::Request<hyper::Body> + Send>>,
+    request_modifier: Option<Box<dyn FnOnce(http::request::Builder) -> hyper::Request<BoxBody<hyper::body::Bytes, Infallible>> + Send>>,
 }
 
 impl std::fmt::Debug for RawRequestBuilder<'_> {
@@ -249,14 +249,14 @@ impl<'a> RawRequestBuilder<'a> {
     /// Set a function to modify the request.
     pub fn map_request<F>(&mut self, f: F) -> &mut Self
     where
-        F: FnOnce(http::request::Builder) -> hyper::Request<hyper::Body> + Send + 'static,
+        F: FnOnce(http::request::Builder) -> hyper::Request<BoxBody<hyper::body::Bytes, Infallible>> + Send + 'static,
     {
         self.request_modifier = Some(Box::new(f));
         self
     }
 
     /// Send the constructed request.
-    pub async fn send(self) -> Result<hyper::Response<hyper::Body>, error::CmdError> {
+    pub async fn send(self) -> Result<hyper::Response<hyper::body::Incoming>, error::CmdError> {
         let url = self.url;
         let method = self.method;
         let cookie_url = self.cookie_url;
@@ -266,7 +266,7 @@ impl<'a> RawRequestBuilder<'a> {
         // *current* domain, whereas we want the cookies for `url`'s domain. So, we navigate to the
         // URL in question, fetch its cookies, and then navigate back. *Except* that we can't do
         // that either (what if `url` is some huge file?). So we *actually* navigate to some weird
-        // url that's unlikely to exist on the target doamin, and which won't resolve into the
+        // url that's unlikely to exist on the target domain, and which won't resolve into the
         // actual content, but will still give the same cookies.
         //
         // The fact that cookies can have /path and security constraints makes this even more of a
@@ -345,7 +345,7 @@ impl<'a> RawRequestBuilder<'a> {
         let req = if let Some(modifier) = request_modifier {
             modifier(req)
         } else {
-            req.body(hyper::Body::empty()).unwrap()
+            req.body(BoxBody::new(http_body_util::Empty::new())).unwrap()
         };
         
         let (tx, rx) = oneshot::channel();
@@ -1095,10 +1095,14 @@ impl Client {
         &self,
         method: Method,
         url: &str,
-    ) -> Result<hyper::Response<hyper::Body>, error::CmdError> {
+    ) -> Result<hyper::Response<hyper::body::Incoming>, error::CmdError> {
         let mut builder = self.raw_request();
-        builder.method(method).url(url);
-        builder.send().await
+        builder
+            .method(method)
+            .url(url);
+        builder
+            .send()
+            .await
     }
 
     /// Build and issue an HTTP request to the given `url` with all the same cookies as the current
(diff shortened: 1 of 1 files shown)


jonhoo (changes requested):

jonhoo:
Thanks for taking this on, and sorry it ended up landing right in the middle of the hyper 1.0 change :sweat_smile:

src/client.rs
```diff
@@ -162,6 +162,127 @@ impl Client {
...
+    
+        let (cookies, ua) = if let Some(cookie_url) = cookie_url {
+            self.client.goto(&cookie_url).await?;
+            let cookies = self.client.issue(WebDriverCommand::GetCookies).await?;
+            self.client.back().await?;
+            if !cookies.is_array() {
+                return Err(error::CmdError::NotW3C(cookies));
+            }
+            let ua = self.client.get_ua().await?;
+            let jar = cookies.as_array().unwrap()
```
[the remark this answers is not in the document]
surajk-m: Yeah, the end result is same. The approach should be more consistent and verbose with the pattern used in `with_raw_client_for`. I've reverted to how you did it to ensure consistency in checks, validating each cookie as an object and that it contains the necessary `name` and `value` fields.

src/client.rs
```diff
@@ -162,6 +162,127 @@ impl Client {
...
     }
+
+    /// Create a new raw request builder.
+    pub fn raw_request(&self) -> RawRequestBuilder<'_> {
+        RawRequestBuilder::new(self)
+    }
+}
+
+/// A builder for constructing raw HTTP requests with optional cookies.
+pub struct RawRequestBuilder<'a> {
```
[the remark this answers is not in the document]
surajk-m: I've refactored `raw_client_for` and `with_raw_client_for` to use `RawRequestBuilder`

src/client.rs
```diff
@@ -162,6 +162,127 @@ impl Client {
         self.issue(Cmd::Persist).await?;
         Ok(())
     }
+
+    /// Create a new raw request builder.
```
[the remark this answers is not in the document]
surajk-m: Thanks, I've updated the description.

src/client.rs
```diff
@@ -162,6 +162,127 @@ impl Client {
...
+        self
+    }
+
+    /// Set the URL for the request.
+    pub fn url(&mut self, url: &str) -> &mut Self {
+        self.url = url.to_string();
+        self
+    }
+
+    /// Set the URL for retrieving cookies.
```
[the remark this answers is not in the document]
surajk-m: Sure, I've added more details and included links to #148 and w3c/webdriver#1238 for context to understand the `WebDriver` limitation and its necessity.

src/client.rs
```diff
@@ -162,6 +162,127 @@ impl Client {
...
+        self.request_modifier = Some(Box::new(f));
+        self
+    }
+
+    /// Send the constructed request.
+    pub async fn send(self) -> Result<hyper::Response<hyper::Body>, error::CmdError> {
+        let url = self.url.clone();
+        let method = self.method.clone();
+        let cookie_url = self.cookie_url.clone();
+        let request_modifier = self.request_modifier.unwrap_or(Box::new(|req| req.body(hyper::Body::empty()).unwrap()));
```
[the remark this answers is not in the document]
surajk-m: Thanks, the updated code is now aligned with `Hyper 1.0` types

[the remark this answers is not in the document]
surajk-m: I've updated it to use `Option` with `if let Some` around the call.

src/client.rs
```diff
@@ -162,6 +162,127 @@ impl Client {
...
+    {
+        self.request_modifier = Some(Box::new(f));
+        self
+    }
+
+    /// Send the constructed request.
+    pub async fn send(self) -> Result<hyper::Response<hyper::Body>, error::CmdError> {
+        let url = self.url.clone();
+        let method = self.method.clone();
+        let cookie_url = self.cookie_url.clone();
```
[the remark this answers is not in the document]
surajk-m: Removed the unnecessary `.clone()` calls.

src/client.rs
```diff
@@ -164,6 +164,199 @@ impl Client {
...
+        //
+        //    example.com/download/some_identifier/ignored_filename_just_for_show
+        //
+        // Imagine if a cookie is set with path=/download/some_identifier. How do we get that
+        // cookie without triggering a request for the (large) file? I don't know. Hence: TODO.
+        //
+        // Retrieve cookies and User-Agent from the WebDriver session if a cookie URL is provided.
+        let (cookies, ua) = if let Some(cookie_url) = cookie_url {
+            // Navigate to the cookie URL to retrieve cookies.
+            self.client.goto(&cookie_url).await?;
```
jonhoo, line 283: The old code would `.join` the current URL with the cookie url — I feel like we may want to do that here as well so that users can provide a relative URL to `.cookie_url` instead of an absolute one?

src/client.rs
```diff
@@ -164,6 +164,199 @@ impl Client {
...
+}
+
+impl<'a> RawRequestBuilder<'a> {
+    /// Create a new raw request builder.
+    pub fn new(client: &'a Client) -> Self {
+        RawRequestBuilder {
+            client,
+            method: Method::GET,
+            url: String::new(),
+            cookie_url: None,
```
jonhoo, line 210: This should probably default to `"/please_give_me_your_cookies"` to maintain parity with the current implementation. And then maybe we add a way to opt _out_ of the cookie navigation?

src/client.rs
```diff
@@ -164,6 +164,199 @@ impl Client {
...
+        // Retrieve cookies and User-Agent from the WebDriver session if a cookie URL is provided.
+        let (cookies, ua) = if let Some(cookie_url) = cookie_url {
+            // Navigate to the cookie URL to retrieve cookies.
+            self.client.goto(&cookie_url).await?;
+            let cookies = self.client.issue(WebDriverCommand::GetCookies).await?;
+            self.client.back().await?;
+            if !cookies.is_array() {
+                return Err(error::CmdError::NotW3C(cookies));
+            }
+            let ua = self.client.get_ua().await?;
```
jonhoo, line 289: I think this should be outside the `if let` since it doesn't have anything to do with the cookies.

src/client.rs
```diff
@@ -924,91 +1119,16 @@ impl Client {
     where
         F: FnOnce(
             http::request::Builder,
-        ) -> hyper::Request<BoxBody<hyper::body::Bytes, Infallible>>,
+        ) -> hyper::Request<BoxBody<hyper::body::Bytes, Infallible>> + Send + 'static,
```
jonhoo, line 1122: I don't think these extra bounds are necessary?

jonhoo (changes requested):
Excellent — just a few more tweaks now and I think we're there!

src/client.rs
surajk-m: Yes, I've updated the code to `.join` the current URL with the `cookie url`.

src/client.rs
surajk-m: I've added an option, we can now choose to skip the cookie navigation process.

src/client.rs
surajk-m: when running tests without `Send` bound, I get the following error -

https://github.com/jonhoo/fantoccini/blob/bce77b0d1a871ce6dafc585e3045e244a79b6218/tests/common.rs#L144  

cargotest

jonhoo: Ah, that's because of
https://github.com/jonhoo/fantoccini/blob/5423b752d9e3b42720153b52621990e190544227/src/client.rs#L188

I think what we'll want to do here is to make the `RawRequestBuilder` generic over `<F>` and store a `Option<F>` instead of `Option<Box<dyn ...>>`. In fact, you could probably get rid of the `Option` as well, and make it:

```rust
struct RawRequestBuilder<'a, F = fn(http::request::Builder) -> hyper::Request<BoxBody<hyper::body::Bytes, Infallible>>> {
    // ...
    request_modifier: F
}

fn empty_body(req: http::request::Builder) -> hyper::Request<BoxBody<hyper::body::Bytes, Infallible>> {
    req.body(BoxBody::new(http_body_util::Empty::new()))
        .unwrap()
}

impl<'a> RawRequestBuilder<'a> {
    /// Create a new raw request builder.
    pub fn new(client: &'a Client) -> Self {
        RawRequestBuilder {
            client,
            // ...
            request_modifier: empty_body,
        }
    }
}

impl<'a, F1> RawRequestBuilder<'a, F1> {
   pub fn map_request<F2>(self, f: F2) -> RawRequestBuilder<'a, F2>
    where
        F: FnOnce(http::request::Builder) -> hyper::Request<BoxBody<hyper::body::Bytes, Infallible>>,
    {
        RawRequestBuilder {
            client: self.client,
            // ...
            request_modifier: f
        }
    }
}
```

Then in `send` you would just unconditionally call `self.map_request`.

jonhoo: By the way, when including output text and such, prefer using [fenced code blocks](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-and-highlighting-code-blocks) to screenshots :)

src/client.rs
jonhoo: I think you could probably get rid of the `bool` and instead have `skip_cookie_navigation` just set `.cookie_url = None`, no? That way you also avoid the extra nested conditional you now have in `send()`.

jonhoo:
It's actually better if you don't force-push while we're still iterating, since the force pushes make it much harder to track exactly what's changed since last time I looked. We can clean up the history at the very end :)

src/client.rs
surajk-m: Yes, the extra `bool` isn't needed. Setting `cookie_url` to `None` works directly to `skip`.

src/client.rs
surajk-m: Thanks, The refactor makes the code more cleaner and efficient.

surajk-m:
Sure, I'll avoid that. Forgot to `rebase` earlier before pushing.

jonhoo:
I made a change so that the various methods are now available for all `<F>`, but otherwise this now looks good, thank you!

jonhoo (approved):

jonhoo:
Released in 0.21.2 :tada:

PR #261, opened 2024-07-15
chromedriver flags that make ci happy eveywhere

dten:
after the stream on the July 14th I tried to dig into why windows selenium is not happy

I don't understand it really but these flags work everywhere on CI, the old flags would fail pretty much every time otherwise


jonhoo:
Fascinating, thank you! May be worth posting a message to https://issues.chromium.org/issues/42323434 since folks there are running into these timeouts too.

jonhoo (approved):

PR #265, opened 2024-08-02
README: fix Method type

cbiffle:
From the README I'm guessing that Method used to be re-exported by fantoccini; that's not the case in the current code. Fixed it to explicitly refer to http::Method.


jonhoo (approved):
Thanks!

PR #275, opened 2024-09-21
fix: Encapsulate `Wcmd` within a private wrapper

### LINKED ISSUE #225: webdriver might be part of public API
I was watching your most recent stream where you mentioned that webdriver isn't part of fantoccini's public API. I have a little experimental tool called [cargo-public-api-crates](https://github.com/davidpdrsn/cargo-public-api-crates) which can find crates your public API, so naturally, I ran it.

It found this for webdriver:

```
webdriver
├── webdriver::command::WebDriverCommand
│   └── src/session.rs:38:0
└── webdriver::command::VoidWebDriverExtensionCommand
    └── src/session.rs:38:0
```

That is [this impl](https://github.com/jonhoo/fantoccini/blob/main/src/session.rs#L38) which does appear to contain types from webdriver.

I'm wondering if this is something you're aware of and if its a false positive in my tool, or if this impl actually does leak the webdriver version.

surajk-m:
Closes #225 

The change encapsulates `Wcmd` within a private wrapper.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,6 @@ license = "MIT OR Apache-2.0"
 default = ["native-tls"]
 native-tls = ["hyper-tls", "openssl"]
 rustls-tls = ["hyper-rustls"]
-test_helpers = []
 
 [dependencies]
 webdriver = { version = "0.50", default-features = false }
@@ -48,7 +47,6 @@ tokio = { version = "1", features = ["full"] }
 hyper = { version = "1.1.0", features = ["server"] }
 hyper-util = { version = "0.1.3", features = ["server", "http1"] }
 serial_test = "3.0"
-fantoccini = { path = "../fantoccini", features = ["test_helpers"] }
 
 # for minimal-versions
 [target.'cfg(any())'.dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,6 +22,7 @@ license = "MIT OR Apache-2.0"
 default = ["native-tls"]
 native-tls = ["hyper-tls", "openssl"]
 rustls-tls = ["hyper-rustls"]
+test_helpers = []
 
 [dependencies]
 webdriver = { version = "0.50", default-features = false }
@@ -47,6 +48,7 @@ tokio = { version = "1", features = ["full"] }
 hyper = { version = "1.1.0", features = ["server"] }
 hyper-util = { version = "0.1.3", features = ["server", "http1"] }
 serial_test = "3.0"
+fantoccini = { path = "../fantoccini", features = ["test_helpers"] }
 
 # for minimal-versions
 [target.'cfg(any())'.dependencies]
--- a/src/session.rs
+++ b/src/session.rs
@@ -336,10 +336,11 @@ impl Client {
     }
 
     /// Issue the specified [`WebDriverCompatibleCommand`] to the WebDriver instance.
-    pub async fn issue_cmd(
-        &self,
-        cmd: impl Into<Box<dyn WebDriverCompatibleCommand + Send + 'static>>,
-    ) -> Result<Json, error::CmdError> {
+    pub async fn issue_cmd<C, CC>(&self, cmd: C) -> Result<Json, error::CmdError>
+    where
+        C: Into<Box<CC>>,
+        CC: WebDriverCompatibleCommand + Send + 'static,
+    {
         self.issue(Cmd::WebDriver(cmd.into())).await
     }
 
--- a/src/session.rs
+++ b/src/session.rs
@@ -26,6 +26,10 @@ type Ack = oneshot::Sender<Result<Json, error::CmdError>>;
 
 type Wcmd = WebDriverCommand<webdriver::command::VoidWebDriverExtensionCommand>;
 
+// Wrapper to hide Wcmd from the public API.
+#[derive(Debug)]
+struct WcmdWrapper(Wcmd);
+
 #[allow(clippy::large_enum_variant)]
 #[derive(Debug)]
 pub(crate) enum Cmd {
@@ -43,7 +47,7 @@ pub(crate) enum Cmd {
     WebDriver(Box<dyn WebDriverCompatibleCommand + Send>),
 }
 
-impl WebDriverCompatibleCommand for Wcmd {
+impl WebDriverCompatibleCommand for WcmdWrapper {
     /// Helper for determining what URL endpoint to use for various requests.
     ///
     /// This mapping is essentially that of <https://www.w3.org/TR/webdriver/#list-of-endpoints>.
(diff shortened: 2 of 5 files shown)


src/session.rs
```diff
@@ -332,11 +336,8 @@ impl Client {
     }
 
     /// Issue the specified [`WebDriverCompatibleCommand`] to the WebDriver instance.
-    pub async fn issue_cmd(
-        &self,
-        cmd: impl WebDriverCompatibleCommand + Send + 'static,
-    ) -> Result<Json, error::CmdError> {
-        self.issue(Cmd::WebDriver(Box::new(cmd))).await
+    pub async fn issue_cmd(&self, cmd: Wcmd) -> Result<Json, error::CmdError> {
```
jonhoo, line 339: I think this specifically needs to _not_ change, otherwise we're putting `Wcmd` back into the public API here.

jonhoo (changes requested):
Thanks for taking this on!

src/session.rs
surajk-m: Apologies for the oversight. I'll revise it to use the `WebDriverCompatibleCommand` trait.

Cargo.toml
```diff
@@ -47,6 +48,7 @@ tokio = { version = "1", features = ["full"] }
 hyper = { version = "1.1.0", features = ["server"] }
 hyper-util = { version = "0.1.3", features = ["server", "http1"] }
 serial_test = "3.0"
+fantoccini = { path = "../fantoccini", features = ["test_helpers"] }
```
jonhoo, line 51: This doesn't look right — how can fantoccini depend on itself with a path of `..`?

src/session.rs
```diff
@@ -334,9 +350,9 @@ impl Client {
     /// Issue the specified [`WebDriverCompatibleCommand`] to the WebDriver instance.
     pub async fn issue_cmd(
         &self,
-        cmd: impl WebDriverCompatibleCommand + Send + 'static,
+        cmd: Box<dyn WebDriverCompatibleCommand + Send + 'static>,
```
jonhoo, line 353: How about taking
```suggestion
        cmd: impl Into<Box<impl WebDriverCompatibleCommand + Send + 'static>>,
```
and then below doing:
```rust
Cmd::WebDriver(Box::from(cmd) as Box<_>)
```
so that callers can also provide a non-boxed type for convenience?

tests/local.rs
```diff
@@ -423,9 +423,13 @@ async fn timeouts(c: Client, _: u16) -> Result<(), error::CmdError> {
...
+        .issue_cmd(test_wrap_command(WebDriverCommand::GetTitle))
+        .await?;
     assert_eq!(title.as_str(), Some("Sample Page"));
-    let title = c.issue_cmd(Box::new(WebDriverCommand::GetTitle)).await?;
+    let title = c
+        .issue_cmd(test_wrap_command(WebDriverCommand::GetTitle))
+        .await?;
     assert_eq!(title.as_str(), Some("Sample Page"));
     Ok(())
 }
```
jonhoo, line 435: I think what we're really looking to demonstrate here is that users can provide their own custom commands via fantoccini. That command doesn't even have to come from the `webdriver` crate. So how about we do this:
```rust
struct GetTitle;
impl WebDriverCompatibleCommand for GetTitle {
  // ...
}
```
and then check that it's possible to pass `GetTitle` and `Box<GetTitle>` to `issue_cmd`? Then you can get rid of `test_wrap_command` entirely (and the test helper feature).

Cargo.toml
surajk-m: Thanks for the feedback. I’ve removed the path reference, it was a misconfiguration introduced during testing `test_helpers` feature.

src/session.rs
surajk-m: I agree, I’ve updated `issue_cmd` to use `impl Into<Box<impl WebDriverCompatibleCommand + Send + 'static>>`

tests/local.rs
surajk-m: Thanks, this change removes the need for `test_wrap_command` and demonstrates `issue_cmd` can accept both `boxed` and `non-boxed` types.

tests/local.rs
```diff
@@ -1,16 +1,49 @@
...
+    ) -> Result<url::Url, url::ParseError> {
+        let base = base_url.join(&format!("session/{}/", session_id.unwrap()))?;
+        base.join("title")
+    }
+
+    fn method_and_body(&self, _: &url::Url) -> (http::Method, Option<String>) {
+        (http::Method::GET, None)
+    }
+
+    fn is_new_session(&self) -> bool {
```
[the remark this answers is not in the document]
surajk-m: No, I'll make the update in the next commit.

tests/local.rs
```diff
@@ -1,16 +1,49 @@
...
+        false
+    }
+
+    fn is_legacy(&self) -> bool {
+        false
+    }
+}
+
+// Implement `Into<Box<dyn WebDriverCompatibleCommand + Send>>` for `GetTitle`
+impl Into<Box<dyn WebDriverCompatibleCommand + Send>> for GetTitle {
```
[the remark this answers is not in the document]
jonhoo: Ah, that just means you'll need to use "proper" generics instead:
```rust
<C, CC> where C: Into<Box<CC>>, CC: WebDriverCompatibleCommand + Send + 'static
```

jonhoo:
Just a heads up that I'll be on holiday for the next three weeks, so won't have a chance to land this until I get back :)

tests/local.rs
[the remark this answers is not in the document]
surajk-m: Thanks, got it. Using `<C, CC>` with `C: Into<Box<CC>>` and `CC: WebDriverCompatibleCommand + Send + 'static` works better. The only overhead is the conversion via `.into()`, which is a bit more than a direct `Box::new` call if the type isn’t already boxed. If we only prefer non-boxed type as input, should we manually box `cmd` with `Box::new(cmd) as Box<_>` as you suggested?

surajk-m:
No worries! Wishing you a great holiday. We’ll pick this up when you're back. 😊 Safe travels!

tests/local.rs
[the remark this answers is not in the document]
jonhoo: I don't think `Box::from` is meaningfully more overhead than `Box::new` in this context, so let's keep it as is :+1: 

Could you please double-check that it's still _possible_ pass a `Box` to `issue_cmd` though, just as a sanity check? Should just be a matter of checking that the test still compiles when you pass `Box::new(GetTitle)`.

jonhoo (approved):
Sorry for the delay here! I think this is now nearly ready, just two things I'd like you to test. One is the `Box::new(GetTitle)` from the open thread, and the other is to re-run the tool from https://github.com/jonhoo/fantoccini/issues/225 and check that `webdriver` indeed no longer appears in the public API. Once those checks are both good, and the conflicts resolved, then we can merge and release :tada:

PR #276, opened 2024-09-25
feat: Add reconnection capability for persisted WebDriver sessions

### LINKED ISSUE #57: Add a way to reconnect to an existing session

surajk-m:
Closes #57   

The `with_existing_session` method introduces the ability to reconnect to an already established WebDriver session using its `session ID`. This enables reuse of an active session without having to create a new one, which is particularly beneficial in scenarios such as process restarts or unexpected disconnections.

### DIFF
--- a/src/session.rs
+++ b/src/session.rs
@@ -651,7 +651,6 @@ where
         client: hyper_util::client::legacy::Client<C, BoxBody<hyper::body::Bytes, Infallible>>,
         wdb: url::Url,
         session_id: Option<&str>,
-        _cap: Option<webdriver::capabilities::Capabilities>,
     ) -> Result<Client, error::NewSessionError> {
         // We're going to need a channel for sending requests to the WebDriver host
         let (tx, rx) = mpsc::unbounded_channel();
@@ -665,7 +664,7 @@ where
 
         // now that the session is running, let's do the handshake
         Ok(Client {
-            tx: tx.clone(),
+            tx,
             is_legacy: false,
             new_session_response: None,
         })
@@ -696,7 +695,7 @@ where
                 .insert("w3c".to_string(), Json::from(true));
         }
 
-        let mut client = Self::setup_session(client, wdb, None, Some(cap.clone())).await?;
+        let mut client = Self::setup_session(client, wdb, None).await?;
 
         let session_config = webdriver::capabilities::SpecNewSessionParameters {
             alwaysMatch: cap.clone(),
@@ -757,7 +756,6 @@ where
                     .await?;
 
                 client.is_legacy = true;
-                client.new_session_response = None;
                 Ok(client)
             }
             Err(e) => Err(e),
--- a/src/session.rs
+++ b/src/session.rs
@@ -535,6 +535,24 @@ impl<C> Session<C>
 where
     C: connect::Connect + Unpin + 'static + Clone + Send + Sync,
 {
+    fn new(
+        rx: mpsc::UnboundedReceiver<Task>,
+        client: hyper_util::client::legacy::Client<C, BoxBody<hyper::body::Bytes, Infallible>>,
+        wdb_url: url::Url,
+        session_id: &str,
+    ) -> Self {
+        Session {
+            ongoing: Ongoing::None,
+            rx,
+            client,
+            wdb: wdb_url,
+            session: Some(session_id.to_string()),
+            is_legacy: false,
+            ua: None,
+            persist: false,
+        }
+    }
+
     fn shutdown(&mut self, ack: Option<Ack>) {
         // session was not created
         if self.session.is_none() {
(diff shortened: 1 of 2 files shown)


src/session.rs
```diff
@@ -535,6 +535,24 @@ impl<C> Session<C>
 where
     C: connect::Connect + Unpin + 'static + Clone + Send + Sync,
 {
+    fn new(
+        rx: mpsc::UnboundedReceiver<Task>,
+        client: hyper_util::client::legacy::Client<C, BoxBody<hyper::body::Bytes, Infallible>>,
+        wdb_url: url::Url,
+        session_id: &str,
```
jonhoo, line 542: How about making this take `Option<impl Into<String>>` and then using `::new` for constructing the `Session` in both the arms of `setup_session`?

src/session.rs
```diff
@@ -607,45 +625,70 @@ where
...
+        cap: &webdriver::capabilities::Capabilities,
+        connector: C,
+    ) -> Result<Client, error::NewSessionError> {
+        let (client, wdb) = Self::create_client_and_parse_url(webdriver, connector).await?;
+        let mut cap = cap.to_owned();
 
-        // Create a new session for this client
-        // https://www.w3.org/TR/webdriver/#dfn-new-session
-        // https://www.w3.org/TR/webdriver/#capabilities
-        //  - we want the browser to wait for the page to load
```
jonhoo, line 648: I think this last line of the comment should stay here as it's specifically related to setting `pageLoadStrategy`.

src/session.rs
```diff
@@ -607,45 +625,70 @@ where
...
-            wdb,
-            session: None,
-            is_legacy: false,
-            ua: None,
-            persist: false,
-        });
+        if let Some(session_id) = session_id {
+            // Reconnect to an existing session
+            tokio::spawn(Session::new(rx, client, wdb, session_id));
+        } else if let Some(_capabilities) = cap {
```
jonhoo, line 662: Why guard this branch behind `cap.is_some()`?

src/session.rs
```diff
@@ -671,7 +720,7 @@ where
             .await
         {
             Ok(new_session_response) => Ok(Client {
-                tx,
+                tx: client.tx,
                 is_legacy: false,
                 new_session_response: Some(wd::NewSessionResponse::from_wd(new_session_response)),
             }),
```
jonhoo, line 726: I think instead of creating a new `Client` here, you can just do `client.new_session_response = Some(..)` and then return `client`.

src/session.rs
```diff
@@ -718,7 +767,7 @@ where
                     .await?;
 
                 Ok(Client {
-                    tx,
+                    tx: client.tx,
                     is_legacy: true,
                     new_session_response: None,
                 })
```
jonhoo, line 773: Similarly here I think you can set `client.is_legacy = true` and return `client`.

jonhoo (changes requested):
This is a great addition, thank you! Just a few smaller bits to tidy up.

src/session.rs
surajk-m: Thanks, revised `::new` to accept `Option<impl Into<String>>`, ensuring it can handle both new and existing sessions seamlessly within `setup_session`.

src/session.rs
surajk-m: Got it, I've removed the check.

src/session.rs
surajk-m: Updated it to set `client.new_session_response` directly before returning.

src/session.rs
```diff
@@ -607,40 +625,59 @@ where
...
 
-        let mut cap = cap.to_owned();
+        Ok((client, wdb))
+    }
+
+    pub(crate) async fn setup_session(
+        client: hyper_util::client::legacy::Client<C, BoxBody<hyper::body::Bytes, Infallible>>,
+        wdb: url::Url,
+        session_id: Option<&str>,
+        _cap: Option<webdriver::capabilities::Capabilities>,
```
jonhoo, line 654: Can we just remove this argument instead?

src/session.rs
```diff
@@ -607,40 +625,59 @@ where
...
-            ua: None,
-            persist: false,
-        });
+            session_id.map(|id| id.to_string()),
+        ));
 
         // now that the session is running, let's do the handshake
-        let client = Client {
+        Ok(Client {
             tx: tx.clone(),
```
jonhoo, line 668: I don't think this needs to clone any more?

src/session.rs
```diff
@@ -717,11 +756,9 @@ where
                     .map(Self::map_handshake_response)
                     .await?;
 
-                Ok(Client {
-                    tx,
-                    is_legacy: true,
-                    new_session_response: None,
-                })
+                client.is_legacy = true;
+                client.new_session_response = None;
```
jonhoo, line 760: I believe this is already set to `None`, so we don't need to set it again?

jonhoo (changes requested):
Looking quite good now, really just nits left :)

src/session.rs
surajk-m: Thanks, I've updated it now.

PR #280, opened 2024-11-19
chore: add update to clientbuilder doc

### LINKED ISSUE #279: Documentation update
found this mentioned somewhere in issues thinking that it was a fantoccini problem but turns out it was a geckodriver issue.

> Client::new should not fail just because another session is already open. I believe this is only the case for geckodriver, and I believe it should be considered a bug there. Works fine with chromedriver for example (this is why the firefox tests in fantoccini all have #[serial]).

If you could add this caveat in clientbuilder connect it would be very helpful. 

Willing to do this myself if you are willing to accept the PR.

hhamud:
* Closes #279


jonhoo (approved):
Perfect, thanks!

PR #282, opened 2025-01-01
fix(client): new_with_connector: correctly setup session

### LINKED ISSUE #281: `basic` Example Fails Due to `Option::unwrap()` on a `None` Value
When running the `basic` example in this project, the execution fails with the following error:

```
thread 'tokio-runtime-worker' panicked at src/session.rs:63:80:
called `Option::unwrap()` on a `None` value
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Error: Lost(Custom { kind: BrokenPipe, error: "WebDriver session was closed while waiting" })
```

#### Steps to Reproduce

1. Clone the `fantoccini` repository.
2. Setup webdriver as per the instruction in `examples/basic.rs`.
3. `cargo run --example basic`

#### Root Cause

After investigating the code, I found that the issue is caused by the `ClientBuilder::connect` method.

https://github.com/jonhoo/fantoccini/blob/452cda5e243ca12dd3412b9b7a23fcce667948e3/src/lib.rs#L255-L261

If capacilities are set, the client will call `Client::with_capabilities_and_connector`, and finally issue a new session id here. Otherwise, `Client::new_with_connector` won't issue a session id.

https://github.com/jonhoo/fantoccini/blob/452cda5e243ca12dd3412b9b7a23fcce667948e3/src/session.rs#L704-L707

#### Possible Solution

Maybe we can use an empty capability as default value? Then in both case we can call `Client::with_capabilities_and_connector` and drop `Client::new_with_connector`.

Bogay:
This change ensures that we issue a new session when we create a new client.

Close #281

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -67,8 +67,7 @@ impl Client {
     where
         C: connect::Connect + Unpin + 'static + Clone + Send + Sync,
     {
-        let (client, wdb) = Session::create_client_and_parse_url(webdriver, connector).await?;
-        Session::setup_session(client, wdb, None).await
+        Session::with_capabilities_and_connector(webdriver, &Default::default(), connector).await
     }
 
     /// Reconnect to a previously established WebDriver session using its ID.


jonhoo (approved):
Perfect, thank you!

jonhoo:
Published as 0.21.4 :tada:

PR #283, opened 2025-01-07
fix: clean up manual JSON string creation

surajk-m:
Clean up manual JSON string construction in `WebDriverCommand::NewSession`.


jonhoo:
Decided to go a very slightly different route that avoids the `if let Some`, but good call-out!
https://github.com/jonhoo/fantoccini/pull/286

PR #285, opened 2025-02-07
feat(client): implement printing pages as PDF

manuelpelloni:
This implements print page to PDF, which can for example be used to generate PDF documents from HTML + CSS templates.
I found the `webdriver` API really unforgiving, so I wrapped everything in constructors and tried to check as much as possible ahead of time.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -118,6 +118,9 @@ pub enum CmdError {
     /// Could not decode a base64 image
     ImageDecodeError(base64::DecodeError),
 
+    /// Could not decode a base64 PDF.
+    PdfDecodeError(base64::DecodeError),
+
     /// Timeout of a wait condition.
     ///
     /// When waiting for a for a condition using [`Client::wait`](crate::Client::wait), any of the
@@ -190,6 +193,7 @@ impl Error for CmdError {
             CmdError::NotW3C(..) => "webdriver returned non-conforming response",
             CmdError::InvalidArgument(..) => "invalid argument provided",
             CmdError::ImageDecodeError(..) => "error decoding image",
+            CmdError::PdfDecodeError(..) => "error decoding PDF",
             CmdError::WaitTimeout => "timeout waiting on condition",
         }
     }
@@ -202,7 +206,7 @@ impl Error for CmdError {
             CmdError::FailedC(ref e) => Some(e),
             CmdError::Lost(ref e) => Some(e),
             CmdError::Json(ref e) => Some(e),
-            CmdError::ImageDecodeError(ref e) => Some(e),
+            CmdError::ImageDecodeError(ref e) | CmdError::PdfDecodeError(ref e) => Some(e),
             CmdError::NotJson(_)
             | CmdError::NotW3C(_)
             | CmdError::InvalidArgument(..)
@@ -224,7 +228,9 @@ impl fmt::Display for CmdError {
             CmdError::NotJson(ref e) => write!(f, "{}", e),
             CmdError::Json(ref e) => write!(f, "{}", e),
             CmdError::NotW3C(ref e) => write!(f, "{:?}", e),
-            CmdError::ImageDecodeError(ref e) => write!(f, "{:?}", e),
+            CmdError::ImageDecodeError(ref e) | CmdError::PdfDecodeError(ref e) => {
+                write!(f, "{:?}", e)
+            }
             CmdError::InvalidArgument(ref arg, ref msg) => {
                 write!(f, "Invalid argument `{}`: {}", arg, msg)
             }
--- a/src/wd.rs
+++ b/src/wd.rs
@@ -1,17 +1,16 @@
 //! WebDriver types and declarations.
-
 use crate::error;
 #[cfg(doc)]
 use crate::Client;
 use http::Method;
 use serde::{Deserialize, Serialize};
-use std::borrow::Cow;
 use std::convert::TryFrom;
 use std::fmt;
 use std::fmt::Debug;
 use std::time::Duration;
+use std::{borrow::Cow, ops::RangeInclusive};
 use url::{ParseError, Url};
-use webdriver::command::TimeoutsParameters;
+use webdriver::command::{PrintParameters, TimeoutsParameters};
 
 /// A command that can be sent to the WebDriver.
 ///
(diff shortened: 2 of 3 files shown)


src/error.rs
```diff
@@ -118,6 +118,9 @@ pub enum CmdError {
     /// Could not decode a base64 image
     ImageDecodeError(base64::DecodeError),
 
+    /// Could not decode a base64 PDF.
+    PdfDecodeError(base64::DecodeError),
```
jonhoo, line 122: Ooof, this is a breaking change. I thought we had already marked this `enum` as `#[non_exhaustive]`, but apparently not :disappointed: Would you mind adding that annotation as well?

src/wd.rs
```diff
@@ -373,3 +372,318 @@ impl NewSessionResponse {
...
+    }
+}
+
+impl PrintConfigurationBuilder {
+    /// Builds the [`PrintConfiguration`].
+    ///
+    /// Returns None if:
+    ///  - the scale, the margins or the size are infinite, NaN or negative
+    ///  - the margins overflow the size (e.g. margins.left + margins.right >= page.width)
+    pub fn build(self) -> Option<PrintConfiguration> {
```
jonhoo, line 408: Should we make this return a `Result` instead of an `Option` to indicate _what_ went wrong?

src/wd.rs
```diff
@@ -373,3 +372,318 @@ impl NewSessionResponse {
...
+    pub fn scale(mut self, scale: f64) -> Self {
+        self.scale = scale;
+
+        self
+    }
+
+    /// Sets whether or not to print the backgrounds of the page.
+    ///
+    /// Default: [`PrintBackground::Exclude`].
+    pub fn background(mut self, background: PrintBackground) -> Self {
```
jonhoo, line 463: I'm okay with this one being a `bool` instead. Or are you expecting this to be a non-binary choice in the future?

src/wd.rs
```diff
@@ -373,3 +372,318 @@ impl NewSessionResponse {
...
+        self.background = background;
+
+        self
+    }
+
+    /// Sets the size of the printed page.
+    ///
+    /// Default: `21.59x27.79 cm`.
```
jonhoo, line 471: I'm surprised it's not legal to submit `None` for this, but I guess the spec requires a size? Also, assuming this a standard paper size, can we name it here and below?

src/wd.rs
```diff
@@ -373,3 +372,318 @@ impl NewSessionResponse {
...
+    pub fn size(mut self, size: PrintSize) -> Self {
+        self.size = size;
+
+        self
+    }
+
+    /// Sets the margins of the printed page.
+    ///
+    /// Default: `1x1x1x1 cm`.
```
jonhoo, line 480: What made you decide this should be the default?

src/wd.rs
```diff
@@ -373,3 +372,318 @@ impl NewSessionResponse {
...
+    pub fn margins(mut self, margins: PrintMargins) -> Self {
+        self.margins = margins;
+
+        self
+    }
+
+    /// Sets ranges of pages to print.
+    ///
+    /// An empty `ranges` prints all pages.
+    /// Default: all.
```
jonhoo, line 490:
```suggestion
    /// An empty `ranges` prints all pages, which is the default.
```

jonhoo (changes requested):
Really good PR, thank you! Some smaller notes. I also wonder how we could write a test for this?

src/wd.rs
manuelpelloni: Firefox uses `0.5inch`  (1.27cm)  and Chromium uses `0.4inch` (1.016cm), but the Webdriver default is 1cm.
I preferred to follow the standard since it's close enough to browsers defaults .

For our own use case we set all margins to zero since we are in control of CSS.
The majority of the websites expect to be printed with additional margins.

src/wd.rs
manuelpelloni: Introduced as `PrintConfigurationError`.

I've also  added a new check that prevents the user from using an invalid `PrintSize`.

src/wd.rs
manuelpelloni: Initially I followed the Webdriver default that is the US legal format, which has this dimensions: `21.59x27.94 cm` (This review references dimensions that I misswrote).

I added A4, US legal and US letter constants, that are some of the formats of the browers print,  to have a nice handy API. 

I made A4 the default since in our own use case we have to print bills and in Europe the legal format is the A4 one, but it seems to be [the most common](https://en.wikipedia.org/wiki/Paper_size#A_series).

manuelpelloni:
> Really good PR, thank you! Some smaller notes. I also wonder how we could write a test for this?

I wrote unit tests to cover all the builder error variants and an integration test that checks the magic bytes of the PDF.

src/error.rs
```diff
@@ -285,6 +292,34 @@ impl From<InvalidWindowHandle> for CmdError {
     }
 }
 
+/// Error of attempting to build an invalid [`PrintConfiguration`](crate::wd::PrintConfiguration).
+#[derive(Debug, PartialEq, Eq)]
+pub enum PrintConfigurationError {
```
jonhoo, line 297:
```suggestion
#[non_exhaustive]
pub enum PrintConfigurationError {
```

src/wd.rs
```diff
@@ -373,3 +372,440 @@ impl NewSessionResponse {
         }
     }
 }
+
+/// The builder of [`PrintConfiguration`].
```
jonhoo, line 376: This is now quite a lot of code. Can we move the print stuff to a `mod print` and then re-export just the `pub` types here?

src/wd.rs
```diff
@@ -373,3 +372,440 @@ impl NewSessionResponse {
...
+    fn default() -> Self {
+        PrintConfiguration {
+            orientation: PrintOrientation::default(),
+            scale: 1.0,
+            background: false,
+            size: PrintSize::default(),
+            margins: PrintMargins::default(),
+            page_ranges: Vec::new(),
+            shrink_to_fit: true,
+        }
```
jonhoo, line 558: Instead of repeating all the defaults here, I think we can just

```suggestion
        PrintConfigurationBuilder::default().build().expect("default configuration is buildable")
```

src/wd.rs
```diff
@@ -373,3 +372,440 @@ impl NewSessionResponse {
...
+
+    /// The standard US legal paper size, which has the dimension of `21.59x35.56cm`.
+    pub const US_LEGAL: Self = Self {
+        width: 21.59,
+        height: 35.56,
+    };
+
+    /// The minimum page size allowed by the Webdriver2 standard, which is `2.54/72cm`.
+    pub const MIN: Self = Self {
+        // FIXME: use 2.54/72.0 with MSRV >= 1.82
```
jonhoo, line 611: Why is the MSRV relevant here?

src/wd.rs
```diff
@@ -373,3 +372,440 @@ impl NewSessionResponse {
...
+impl PrintOrientation {
+    pub(crate) fn into_params(self) -> webdriver::command::PrintOrientation {
+        match self {
+            Self::Landscape => webdriver::command::PrintOrientation::Landscape,
+            Self::Portrait => webdriver::command::PrintOrientation::Portrait,
+        }
+    }
+}
+
+/// The size of the printed page in centimeters.
```
jonhoo, line 581: Let's also document the default here.

src/wd.rs
```diff
@@ -373,3 +372,440 @@ impl NewSessionResponse {
...
+
+        if start == end {
+            webdriver::command::PrintPageRange::Integer(start)
+        } else {
+            webdriver::command::PrintPageRange::Range(format!("{start}-{end}"))
+        }
+    }
+}
+
+/// The margins of the printed page in centimeters.
```
jonhoo, line 664: Let's document the default here as well

jonhoo (changes requested):
Nice, we're super close now!

src/wd.rs
manuelpelloni: The MIN value (0.036) is an approximate result of the 2.54/72.0 operation.
Without the approximation, the result would be similar to 0.03527777777777778, but unfortunately constant floating point operations were introduced in [Rust 1.82](https://blog.rust-lang.org/2024/10/17/Rust-1.82.0.html#floating-point-nan-semantics-and-const).

jonhoo: Ohhh, okay, yeah, makes sense, thanks!

jonhoo (approved):
Perfect, thank you! I'm going to hold off on releasing this just until #289 lands (which should be very soon) since it also has breaking changes. Thanks for sticking with it!

jonhoo:
Ah, some of the CI is racy: fixed in https://github.com/jonhoo/fantoccini/pull/293

PR #287, opened 2025-02-16
Lighten dependency on `futures` crates

paolobarbolini:
This lightens the dependency on the `futures` crates by removing the direct dependency on `futures-core`, instead using the `ready!` macro from std, and reducing the features on `futures-util` to just the ones we need.


jonhoo (approved):
Excellent, thank you!

PR #289, opened 2025-03-02
Support querying numeric HTML properties from `elements::Element::prop()`

johnDeSilencio:
## Preface

Hi, Jon! Longtime fan of yours. Your Crust of Rust series was HUGE for me going from a beginner Rust developer to more intermediate (I've got a long way to go before the rank of "master" :smile:). Thank you for all the amazing work you do and how you give back to the community!

## Problem

I'm building [my own personal portfolio website](https://johndesilenc.io). One of the features of the application is a terminal emulator, letting you [run fake commands on the website](https://github.com/johnDeSilencio/johnDeSilencio.github.io/blob/main/src/main.rs#L60-L77).

One of the commands is `clear`, which clears the screen and the history of all previous commands. To write a functional test for this feature, I'd like to be able to tell that my `<div id="previous-commands>` has no children which I check through the [`childElementCount`](https://developer.mozilla.org/en-US/docs/Web/API/Element/childElementCount) property.

However, when I try use [`fantoccini::elements::Element::prop("childElementCount")`](https://docs.rs/fantoccini/latest/fantoccini/elements/struct.Element.html#method.prop), I get the following error:

`...webdriver returned non-conforming response: Number(0)`

2025-03-01-162717_hyprshot

## Proposed Solution

It appears that the `fantoccini::elements::Element::prop` method currently only handles `String`s, `Bool`s, and `Null` values from `serde_json` and returns an error for everything else. This PR also supports matching [`serde_json::Number`](https://docs.rs/serde_json/latest/serde_json/struct.Number.html) to support getting numeric properties like `childElementCount`.

## Tests

I've also added another `assert_eq!()` to the `element_test` function. Let me know if you'd like me to add more test coverage or to find another way to support numeric HTML properties.

## Functional Tests

I used this fork in the E2E tests for my website and now my functional tests pass:

image

### DIFF
--- a/src/elements.rs
+++ b/src/elements.rs
@@ -334,7 +334,7 @@ impl Element {
     #[cfg_attr(docsrs, doc(alias = "outerHTML"))]
     pub async fn html(&self, inner: bool) -> Result<String, error::CmdError> {
         let prop = if inner { "innerHTML" } else { "outerHTML" };
-        Ok(self.prop(prop).await?.unwrap())
+        Ok(self.prop(prop).await?.unwrap().to_string())
     }
 }
 
--- a/src/elements.rs
+++ b/src/elements.rs
@@ -370,7 +370,13 @@ impl Element {
     #[cfg_attr(docsrs, doc(alias = "outerHTML"))]
     pub async fn html(&self, inner: bool) -> Result<String, error::CmdError> {
         let prop = if inner { "innerHTML" } else { "outerHTML" };
-        Ok(self.prop(prop).await?.unwrap().to_string())
+
+        match self.prop(prop).await? {
+            // Requesting `innerHTML` or `outerHTML` should normally return a string
+            Some(Json::String(contents)) => Ok(contents),
+            Some(res) => Err(error::CmdError::NotW3C(res)),
+            _ => Err(error::CmdError::NotW3C(Json::Null)),
+        }
     }
 }
 
--- a/src/elements.rs
+++ b/src/elements.rs
@@ -230,14 +230,11 @@ impl Element {
     ///
     /// [property]: https://www.ecma-international.org/ecma-262/5.1/#sec-8.12.1
     #[cfg_attr(docsrs, doc(alias = "Get Element Property"))]
-    pub async fn prop(&self, prop: &str) -> Result<Option<String>, error::CmdError> {
+    pub async fn prop(&self, prop: &str) -> Result<Option<Json>, error::CmdError> {
         let cmd = WebDriverCommand::GetElementProperty(self.element.clone(), prop.to_string());
         match self.client.issue(cmd).await? {
-            Json::String(v) => Ok(Some(v)),
-            Json::Bool(b) => Ok(Some(b.to_string())),
-            Json::Number(n) => Ok(Some(n.to_string())),
             Json::Null => Ok(None),
-            v => Err(error::CmdError::NotW3C(v)),
+            v => Ok(Some(v)),
         }
     }
 
--- a/src/elements.rs
+++ b/src/elements.rs
@@ -223,12 +223,48 @@ impl Element {
     ///
     /// `Ok(None)` is returned if the element does not have the given property.
     ///
-    /// Boolean properties such as "checked" will be returned as the String "true" or "false".
-    ///
     /// See [13.3 Get Element Property](https://www.w3.org/TR/webdriver1/#get-element-property)
     /// of the WebDriver standard.
     ///
     /// [property]: https://www.ecma-international.org/ecma-262/5.1/#sec-8.12.1
+    ///
+    /// Example:
+    ///
+    /// ```
+    /// # use fantoccini::{ClientBuilder, error, Locator};
+    /// # use serde_json::{json, Number};
+    /// # async fn no_search_results() -> Result<(), error::CmdError> {
+    /// // Define capabilities of client...
+    /// #    let capabilities = json!({})
+    /// #        .as_object()
+    /// #        .unwrap()
+    /// #        .to_owned();
+    /// #
+    /// // Create client...
+    /// # let client = ClientBuilder::native()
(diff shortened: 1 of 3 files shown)


tests/elements.rs
```diff
@@ -48,6 +48,9 @@ async fn element_prop(c: Client, port: u16) -> Result<(), error::CmdError> {
     assert_eq!(elem.prop("id").await?.unwrap(), "checkbox-option-1");
     assert_eq!(elem.prop("checked").await?.unwrap(), "false");
     assert!(elem.attr("invalid-property").await?.is_none());
+
+    let elem = c.find(Locator::Id("content")).await?;
+    assert_eq!(elem.prop("childElementCount").await?.unwrap(), "5");
```
johnDeSilencio, line 53: This is the number of children the `<div>` with ID `content` has in the example HTML page. If the example changes, this test will have to change as well

jonhoo (changes requested):
Thank you^2!

Arguably, `prop` should just return a `serde_json::Value` instead of coercing everything to a string. I'm honestly not quite sure why I made it be string only :thinking: I think it's because [attribute can only return strings](https://dom.spec.whatwg.org/#concept-attribute), and I guess I incorrectly assumed that the same would apply to `prop`. But [property values can be "any ECMAScript language type"](https://262.ecma-international.org/5.1/#sec-8.6.1). Given we're going to have to cut a breaking change for https://github.com/jonhoo/fantoccini/pull/285 anyway, let's maybe just go ahead and change `prop` to return `serde_json::Value` right now?

johnDeSilencio:
Hey, Jon!

I've made the change you suggested so that `prop()` now returns a `serde_json::Value`. I also added a doc comment example to the `prop()` method, but feel free to let me know if you think I should take it out or change it

src/elements.rs
```diff
@@ -337,7 +370,7 @@ impl Element {
     #[cfg_attr(docsrs, doc(alias = "outerHTML"))]
     pub async fn html(&self, inner: bool) -> Result<String, error::CmdError> {
         let prop = if inner { "innerHTML" } else { "outerHTML" };
-        Ok(self.prop(prop).await?.unwrap())
+        Ok(self.prop(prop).await?.unwrap().to_string())
```
jonhoo, line 373: I wonder if we should be a bit more principled here and instead of `.unwrap().to_string()` specifically match for `Some(Json::String` and error anything else?

jonhoo (commented):
That looks way better, thank you! One more question/change and then I think we're good!

jonhoo:
Some of the tests in CI seem unhappy :thinking:

jonhoo:
Ah, some of the CI is just racy: fixed in https://github.com/jonhoo/fantoccini/pull/293

src/elements.rs
```diff
@@ -223,20 +223,54 @@ impl Element {
...
+    ///     .expect("expected childElementCount to be an integer value");
+    ///
+    /// assert_eq!(num_search_matches, 0);
+    /// #
+    /// #   Ok(())
+    /// # };
+    /// ```
     #[cfg_attr(docsrs, doc(alias = "Get Element Property"))]
-    pub async fn prop(&self, prop: &str) -> Result<Option<String>, error::CmdError> {
+    pub async fn prop(&self, prop: &str) -> Result<Option<Json>, error::CmdError> {
```
paolobarbolini, line 269: Drive-by review: wouldn't it be better without the `Option`?

jonhoo: It's a good question. I was thinking about that while reading through. I _think_ it's probably good to encode the knowledge inside `fantoccini` that "no such property" is encoded by spec as `null`, rather than requiring the user to be aware of that?

src/elements.rs
```diff
@@ -336,7 +370,13 @@ impl Element {
     #[cfg_attr(docsrs, doc(alias = "outerHTML"))]
     pub async fn html(&self, inner: bool) -> Result<String, error::CmdError> {
         let prop = if inner { "innerHTML" } else { "outerHTML" };
-        Ok(self.prop(prop).await?.unwrap())
+
+        match self.prop(prop).await? {
+            // Requesting `innerHTML` or `outerHTML` should normally return a string
+            Some(Json::String(contents)) => Ok(contents),
+            Some(res) => Err(error::CmdError::NotW3C(res)),
+            _ => Err(error::CmdError::NotW3C(Json::Null)),
```
johnDeSilencio, line 378: Not sure if this is the most appropriate error variant for the `None` case. Open to suggestions

johnDeSilencio:
The prodigal PR author returns 😅

I've merged up my branch with the most recent work on `main`. I've also made the change to the `Element::html()` method that you suggested. Wasn't sure exactly what variant of `error::CmdError` to return. Feel free to let me know if you think a different variant would be more appropriate!

---

As a side note, I want to apologize for taking so long to get back to you and to this PR. My personal life got kinda hectic, and I let this PR fall through the cracks. I also apologize if my tardiness has delayed the release of 0.23. Thank you for your patience and the gentle reminder last month :smiley: 

Moving forward, I've gotten my schedule mapped out pretty well in self-hosted CalDav calendars (shout-out to [Stalwart](https://stalw.art/docs/collaboration/calendar)), so if I do submit future PRs, the turn-around time should be significantly less than 4 months!

jonhoo (approved):
No worries at all — it happens to all of us from time to time!

As for releasing this, I think I'll then bundle this with https://github.com/jonhoo/fantoccini/pull/275 in a 0.23, but that does mean it may take a little while to get out :)

johnDeSilencio:
> No worries at all — it happens to all of us from time to time!
> 
> As for releasing this, I think I'll then bundle this with https://github.com/jonhoo/fantoccini/pull/275 in a 0.23, but that does mean it may take a little while to get out :)

Thanks for understanding! And yeah, no rush on the release either :)

PR #294, opened 2025-03-22
cargo update

jonhoo:
Breaking change because the webdriver crate now generates floating point numbers for pointer move operations, which are rejected by older versions of Chrome and Firefox. See https://bugzilla.mozilla.org/show_bug.cgi?id=1959464 for details.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -112,7 +112,7 @@ jobs:
     # https://docs.github.com/en/actions/learn-github-actions/contexts#context-availability
     strategy:
       matrix:
-        msrv: ["1.71.1"] # tokio
+        msrv: ["1.82.0"] # tokio
     name: ubuntu / ${{ matrix.msrv }}
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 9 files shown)


jonhoo:
This currently fails because the `webdriver` crate failed to make the `capabilities` field of `NewSessionParameters` `pub` when converting it from an `enum` to a `struct` as part of the removal of legacy support in 0.51. I've reported it to [the mailing list](https://groups.google.com/a/mozilla.org/g/dev-webdriver), but we'll be blocked from upgrading until a fix lands.

jonhoo:
Upstream bug report: https://bugzilla.mozilla.org/show_bug.cgi?id=1955903

jonhoo:
Worth noting that this will be a breaking change, but that's okay since we'll already have to break with #289 and #285.

jonhoo:
Well this is awkward: the current versions of Firefox and Chrome both only allow integer values for the `x` and `y` fields of pointer action move commands, but `webdriver` now encodes them as `f64`. I have followed up upstream in the same bug report.

jonhoo:
minimal-versions in CI will remain broken until webdriver upgrades to a newer version of icu_segmenter (specifically: 2.0) so that they move on from the zerovec version that has this stanza: https://github.com/unicode-org/icu4x/blob/icu%401.5.0/utils/zerovec/Cargo.toml#L33-L38

PR #297, opened 2025-04-14
printsize (fix): swap const width and height for A4

Aprax14:
This fixes the A4 paper size const definition in the printsize struct, where the width and height values were inverted


paolobarbolini:
@jonhoo I see things are stalled in the other PRs. We'd like to start using #285 + the fix in this PR. Could we make a new release with this merged and leave everything else to the future?

